PR #11, opened 2023-12-12
Create the `cloudevent-recorder` module.

mattmoor:
This module makes it easy to configure a recorder service that pulls events of particular types off of a `cloudevent-broker`'s topics, rotates them into GCS, and sets up a BigQuery DTS job to import them into a schematized table of those events.


cloudevent-recorder/bigquery.tf
```diff
@@ -0,0 +1,100 @@
...
+  location             = google_bigquery_dataset.this.location // These must be colocated
+  service_account_name = google_service_account.import-identity.email
+  disabled             = false
+
+  data_source_id         = "google_cloud_storage"
+  schedule               = "every 15 minutes"
+  destination_dataset_id = google_bigquery_dataset.this.dataset_id
+
+  // TODO(mattmoor): Bring back pubsub notification.
+  # notification_pubsub_topic = google_pubsub_topic.bq_notification[each.key].id
```
mattmoor, line 92: cc @k4leung4 

When you are back, we should chase this down.  I dropped it in my prototyping.

PR #14, opened 2023-12-19
add dashboard modules

imjasonh:
This moves modules in from https://github.com/chainguard-dev/terraform-cloudrun-dashboard -- after this is merged we can archive that repo.

The new directory `./dashboard` includes:
- `service` and `job`, which take a multi-regional Cloud Run Service or Job name and presents common metrics for them
- `topic`, which takes a Pub/Sub topic prefix and presents common metrics for its subscribers, assuming they have that topic name as their subscription name prefix
- `tiles`, which provide reusable dashboard tile definitions, which we can extend over time.

It also defines dashboards for the ingress and recorder services, and broker topics.


cloudevent-broker/main.tf
```diff
@@ -13,3 +13,9 @@ resource "google_pubsub_topic" "this" {
   // TODO: Tune this and/or make it configurable?
   message_retention_duration = "600s"
 }
+
+module "dashboard" {
```
mattmoor, line 17: This collides with the invocation in `ingress.tf`, but I'd like to see if we can combine these into a unified broker dashboard using your building blocks 🤩

dashboard/job/variables.tf
```diff
@@ -0,0 +1,9 @@
+variable "project_id" { type = string }
+variable "job_name" { type = string }
+
+variable "alert_policies" {
+  type = map(object({
+    id = string
+  }))
+  default = {}
```
mattmoor, line 8: I've been trying to add detailed comments to the `variables` and `outputs` so our auto-docs are 💯

dashboard/tiles/latency/main.tf
```diff
@@ -0,0 +1,39 @@
+variable "title" { type = string }
+variable "group_by_fields" { default = [] }
+variable "filter" { type = list(string) }
```
mattmoor, line 3: nit: split `variables.tf` and `outputs.tf` 🙏

mattmoor: oh fun, the whole thing is inputs/outputs 🤔

dashboard/job/variables.tf
mattmoor: I've also been trying to add fairly decent usage comments to the `README.md` files.

dashboard/topic/dashboard.tf
```diff
@@ -0,0 +1,76 @@
...
+  primary_reduce  = "REDUCE_NONE"
+}
+
+module "push-latency" {
+  source = "../tiles/latency"
+  title  = "Push latency"
+  filter = [
+    "resource.type=\"pubsub_subscription\"",
+    "metric.type=\"pubsub.googleapis.com/subscription/push_request_latencies\"",
+    "resource.label.\"subscription_id\"=monitoring.regex.full_match(\"${var.topic_prefix}.*\")",
```
mattmoor, line 46: `topic_prefix` seems wrong.  I think you want `subscription_prefix`?

.github/workflows/documentation.yaml
```diff
@@ -15,6 +15,9 @@ jobs:
         - cloudevent-trigger
         - cloudevent-recorder
         - networking
+        - dashboard/service
+        - dashboard/job
+        - dashboard/topic
```
mattmoor, line 20: Do we want the tiles too?

mattmoor: subscription?

imjasonh: Ehhh, maybe. They're so simple I think we'd end up having most of the content of the directories be comments, and generated markdown from comments.

Which maybe isn't a bad thing! I just don't want to have to write it all right now 😆

dashboard/tiles/latency/main.tf
imjasonh: Yeah, these tiles are really simple little things. I could split them up but it's just more tab-flipping when you iterate on them.

dashboard/job/variables.tf
imjasonh: Can I give you an IOU for the alert policy stuff for now? That's my next chunk of work and I might completely redo how it's structured.

.github/workflows/documentation.yaml
mattmoor: LOL, ok.  I've been trying to keep the bar here relatively high, so let's take the AI to chase this once we feel like the shape is something we're really bought into?

imjasonh:
Dropped alert policies for now, I'll readd it later. Also dropped the project_id variable, which was only used to power the url output, which I've also removed.

cloudevent-broker/dashboard.tf
```diff
@@ -0,0 +1,88 @@
+module "sent-events" {
+  source = "../dashboard/tiles/xy"
+  title  = "Events Sent"
+  filter = [
+    "resource.type=\"pubsub_topic\"",
+    "metric.type=\"pubsub.googleapis.com/topic/send_request_count\"",
```
mattmoor, line 6: We should filter this on the topic name prefix

cloudevent-broker/main.tf
```diff
@@ -13,3 +13,9 @@ resource "google_pubsub_topic" "this" {
   // TODO: Tune this and/or make it configurable?
   message_retention_duration = "600s"
 }
+
+module "topic-dashboard" {
+  source       = "../dashboard/topic"
+  project_id   = var.project_id
+  topic_prefix = var.name
+}
```
mattmoor, line 21: How is this different from `dashboard.tf`? 🤔

.github/workflows/documentation.yaml
imjasonh: re: topic vs subscription, wdyt about `dashboard/subscriptions` (with an `s`)?

The PubSub dashboard is weird because it doesn't just go by the (regionally-unique) name of the thing, it goes by prefixes on (globally-unique) subscription names.

cloudevent-broker/dashboard.tf
mattmoor: See my other comment, but I think you factored this out (with prefix!) and then forgot to clean this up?

cloudevent-broker/main.tf
imjasonh: Oh it's not! This was an artifact of refactoring. Removed.

.github/workflows/documentation.yaml
imjasonh: The sent-events tile for instance relies on the fact that `subscription_prefix` is also a topic name.

dashboard/tiles/alert/main.tf
```diff
@@ -0,0 +1,10 @@
+variable "title" { type = string }
+variable "alert_name" { type = string }
+
+// https://cloud.google.com/monitoring/api/ref_v3/rest/v1/projects.dashboards#AlertChart
+output "tile" {
```
mattmoor, line 5: These are actually widgets

cloudevent-broker/dashboard.tf
mattmoor: oh weird, no.

mattmoor: The "topic" dashboard shows "subscription" stats 🤔

.github/workflows/documentation.yaml
mattmoor: We should be able to filter subscriptions metrics on `topic_id` without assuming the subscription prefix matches.  I was playing with this in metrics explorer earlier.

mattmoor: But let's land the widgets / cloudrun dashboards, and then talk pubsub next

dashboard/topic/dashboard.tf
mattmoor, line 1: drop for now

dashboard/subscription/dashboard.tf
mattmoor, line 1: drop for now

dashboard/subscription/README.md
mattmoor, line 1: drop for now

.github/workflows/documentation.yaml
mattmoor, line 20: drop for now
```suggestion
```

dashboard/widgets/xy/main.tf
```diff
@@ -0,0 +1,46 @@
...
+variable "filter" { type = list(string) }
+variable "plot_type" { default = "LINE" }
+variable "alignment_period" { default = "60s" }
+variable "primary_align" { default = "ALIGN_RATE" }
+variable "primary_reduce" { default = "REDUCE_NONE" }
+variable "secondary_align" { default = "ALIGN_NONE" }
+variable "secondary_reduce" { default = "REDUCE_NONE" }
+
+// https://cloud.google.com/monitoring/api/ref_v3/rest/v1/projects.dashboards#XyChart
+output "tile" {
```
mattmoor, line 12: We should rename the outputs too.

mattmoor (approved):

PR #19, opened 2023-12-20
Add an in-process otel collector

tcnghia:
This will automatically scrape prometheus metrics and send to GMP.


pkg/otel/doc.go
```diff
@@ -0,0 +1,20 @@
+package otel
```
mattmoor, line 1: you want this under the package comment

pkg/otel/components.go
```diff
@@ -0,0 +1,154 @@
...
+	// This is workaround since we don't have a config provider that just take a YAML blob.
+	// Once we have that, we can just pass the default config as a string to the config provider.
+	configFile, err := os.CreateTemp("", "")
+	if err != nil {
+		return nil, err
+	}
+	if configYAML != "" {
+		configYAML = defaultConfig
+	}
+	os.WriteFile(configFile.Name(), []byte(configYAML), 0644)
```
mattmoor, line 71:
```suggestion
	if err := os.WriteFile(configFile.Name(), []byte(configYAML), 0644); err != nil {
		return nil, err
	}
```

pkg/otel/doc.go
```diff
@@ -0,0 +1,20 @@
...
+// collector asynchronously,
+//
+//     shutdown, runErr, err := otel.StartCollectorAsync(ctx)
+//     if err != nil {
+// 	      // handle errors
+//     }
+// 	   defer shutdown()
+//     // ... optionally wait for runErr and process it.
+//
+// The prometheus endpoint is expected to be serving at localhost:9090/metrics.
```
mattmoor, line 20:
```suggestion
// package otel provides an OpenTelemetry Collector with components
// suitable to be used on Cloud Run.
//
// This is based on the sidecar example at
// https://cloud.google.com/run/docs/tutorials/custom-metrics-sidecar,
// but with an in-process collector instead of a sidecar.
//
// The main function to use is StartCollectorAsync, which starts the
// collector asynchronously,
//
//     shutdown, runErr, err := otel.StartCollectorAsync(ctx)
//     if err != nil {
// 	      // handle errors
//     }
// 	   defer shutdown()
//     // ... optionally wait for runErr and process it.
//
// The prometheus endpoint is expected to be serving at localhost:9090/metrics.
package otel
```

pkg/otel/components.go
```diff
@@ -0,0 +1,156 @@
...
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/googlecloudexporter"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/googlemanagedprometheusexporter"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/extension/healthcheckextension"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourcedetectionprocessor"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourceprocessor"
+)
+
+var (
+	//go:embed default.yaml
```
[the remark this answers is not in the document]
tcnghia: Yes, I confirmed that this works on a real service.

mattmoor (approved):

PR #26, opened 2023-12-21
cleanup: detect project ID instead of passing via env

imjasonh:
Cloud Run services know what project they're running in, let's just use that.


imjasonh:
The pubsub client's weird, it requires a `projectID` but it also detects it: https://github.com/googleapis/google-cloud-go/blob/pubsub/v1.33.0/pubsub/pubsub.go#L181C22-L181C22

imjasonh:
Ahh, `pubsub.DetectProjectID`

mattmoor (approved):

PR #27, opened 2023-12-21
support passing alert IDs to service dashboards

imjasonh:
This should produce full-width alert chart widgets for each alert. I might need to iterate on this with changes in tf-prober to see this all working together. It at least doesn't regress for non-alert-having services.

### DIFF
--- a/dashboard/service/dashboard.tf
+++ b/dashboard/service/dashboard.tf
@@ -16,15 +16,24 @@ module "resources" {
   filter = ["resource.type=\"cloud_run_revision\""]
 }
 
+module "alerts" {
+  source = "../sections/alerts"
+  alert  = var.alert
+  title  = "Alert"
+}
+
 module "width" { source = "../sections/width" }
 
 module "layout" {
-  source   = "../sections/layout"
-  sections = [
-    module.logs.section,
-    module.http.section,
-    module.resources.section,
-  ]
+  source = "../sections/layout"
+  sections = concat(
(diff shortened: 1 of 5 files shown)


dashboard/service/dashboard.tf
```diff
@@ -16,15 +16,24 @@ module "resources" {
   filter = ["resource.type=\"cloud_run_revision\""]
 }
 
+module "alerts" {
+  source = "../sections/alerts"
```
mattmoor, line 20:
```suggestion
  count  = var.alert != "" ? 1 : 0
  source = "../sections/alerts"
```

dashboard/service/dashboard.tf
```diff
@@ -16,15 +16,24 @@ module "resources" {
...
 module "layout" {
-  source   = "../sections/layout"
-  sections = [
-    module.logs.section,
-    module.http.section,
-    module.resources.section,
-  ]
+  source = "../sections/layout"
+  sections = concat(
+    var.alert == "" ? [] : [module.alerts.section],
```
mattmoor, line 30:
```suggestion
    var.alert == "" ? [] : [module.alerts[0].section],
```

or
```suggestion
    [for x in module.alerts : x.section],
```

dashboard/service/dashboard.tf
mattmoor: We could also consider making this a list 🤔

mattmoor (approved):

PR #28, opened 2023-12-22
Add outbound HTTP dashboard

tcnghia:
This based on the metrics we export in our metricking transport.

There are some difference with Cloud Run native metrics:
1. we can't add `resource.type` to our metrics. I work around by not using `resource.type` in the filter here.
2. the cloud run native metrics has `service_name` being a `resource.label`, which AFAIK we can't add to our Prometheus metrics. when following the instruction [here](https://cloud.google.com/run/docs/tutorials/custom-metrics-opentelemetry-sidecar) the `service_name` is a `metric.label`. As a result, we will need to add that filter here, unlike the other widgets that can inherit from the Dashboard's. 

For 2. we could also add a template variable at the Dashboard level that use a `metric.label`, but will look a bit confusing as we have two drop downs that look exactly the same. Since we are baking `var.service_name` to the dashboard name anyway, I think we may as well drop the template filter from there.

This is how this change would look like (if the metrics are there)

image

### DIFF
--- a/dashboard/sections/http/main.tf
+++ b/dashboard/sections/http/main.tf
@@ -1,6 +1,7 @@
 variable "title" { type = string }
 variable "filter" { type = list(string) }
 variable "collapsed" { default = false }
+variable "service_name" { type = string }
 
 module "width" { source = "../width" }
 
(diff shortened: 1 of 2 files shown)


dashboard/sections/http/main.tf
```diff
@@ -21,7 +22,42 @@ module "incoming_latency" {
...
+locals {
+  // gmp_filter is a subset of var.filter that does not include the "resource.type" string
+  gmp_filter = [for f in var.filter : f if !strcontains(f, "resource.type")]
+}
+
+output "gmp_filter" {
+  value = local.gmp_filter
+}
+
+output "var_filter" {
```
imjasonh, line 34: We don't need this, since anything that wants to share its value can already just get `.filter`

dashboard/sections/http/main.tf
```diff
@@ -21,7 +22,42 @@ module "incoming_latency" {
...
+  value = var.filter
+}
+
 // TODO(mattmoor): output HTTP charts.
```
imjasonh, line 38: Can we remove this now?

dashboard/sections/http/main.tf
tcnghia: Done

dashboard/sections/http/main.tf
tcnghia: Done

PR #37, opened 2023-12-22
Add gRPC (inbound)

tcnghia:
Currently our gRPC metrics dp not have the right label `service_name` yet. That means we can't have outbound gRPC widgets (which would need to distinguish by caller i.e. `service_name`).

I'm using the `grpc_service_name` variable as a work around here. Will remove it and add outbound gRPC when the metric label is fixed.

### DIFF
--- a/dashboard/sections/grpc/grpc.tf
+++ b/dashboard/sections/grpc/grpc.tf
@@ -1,0 +1,105 @@
+variable "title" { type = string }
+variable "filter" { type = list(string) }
+variable "collapsed" { default = false }
+variable "grpc_service_name" { type = string }
+
+module "width" { source = "../width" }
+
+module "request_count" {
+  source = "../../widgets/xy"
+  title  = "Request count"
+  filter = concat(var.filter, [
+    "metric.type=\"prometheus.googleapis.com/grpc_server_handled_total/counter\"",
+    "metric.label.grpc_service=monitoring.regex.full_match(\"${var.grpc_service_name}.*\")",
+  ])
+  group_by_fields = [
+    "metric.label.\"grpc_service\"",
+    "metric.label.\"grpc_method\"",
+    "metric.label.\"grpc_code\""
+  ]
+  primary_align    = "ALIGN_RATE"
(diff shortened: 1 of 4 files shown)


imjasonh (approved):

dashboard/service/variables.tf
```diff
@@ -14,3 +14,11 @@ variable "alerts" {
   default     = []
 }
 
+# Currently our metrics does not have service_name label: we
+# are working around by specifying the grpc_service name label
+# instead while we fix the metric labeling.
+variable "grpc_service_name" {
+  description = "Name of the GRPC service(s) to monitor"
+  type        = string
+  default     = ""
```
mattmoor, line 23: I don't think we should have a default here

dashboard/sections/grpc/grpc.tf
```diff
@@ -0,0 +1,105 @@
...
+  secondary_align  = "ALIGN_NONE"
+  secondary_reduce = "REDUCE_SUM"
+}
+
+module "outbound_latency" {
+  source = "../../widgets/latency"
+  title  = "Incoming request latency"
+  filter = concat(var.filter, [
+    "metric.type=\"prometheus.googleapis.com/grpc_server_handling_seconds/histogram\"",
+    "metric.label.\"grpc_service\"=monitoring.regex.full_match(\"chainguard.datastore.*\")",
```
mattmoor, line 62: hmm `chainguard.datastore` here seems wrong.

dashboard/sections/grpc/grpc.tf
```diff
@@ -0,0 +1,105 @@
...
+    "metric.label.\"grpc_method\"",
+    "metric.label.\"grpc_code\""
+  ]
+  primary_align    = "ALIGN_RATE"
+  primary_reduce   = "REDUCE_NONE"
+  secondary_align  = "ALIGN_NONE"
+  secondary_reduce = "REDUCE_SUM"
+}
+
+module "outbound_latency" {
```
mattmoor, line 57: These are unused?

PR #42, opened 2023-12-24
otel-collector config to populate FAAS name

tcnghia:
The `gcp` resource detection plugin populates the attribute `faas.name` to Cloud Run Service/Job, and Cloud Function name, which makes it a nice way to standardize on naming ([see](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md)). Using env var is less convenient for this purpose because Cloud Run Service name is `$K_SERVICE` and for Cloud Run Job it's `$CLOUD_RUN_JOB`.

The googlemanagedprometheus exporter consumes `service.name` attribute and create a resource label named `job` with this value when uploading metrics. There is no customization of this. ([see](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/764)).

This PR plumbs `faas.name` into `service.name` so we can get a resource label with the Cloud Run Service or Job name on all our exported metrics. Also dropping `instance id` because this is an unbounded dimension, and even worse when a lot of scaling-to-zero happens.

### DIFF
--- a/otel-collector/cmd/otel-collector/kodata/config.yaml
+++ b/otel-collector/cmd/otel-collector/kodata/config.yaml
@@ -7,6 +7,8 @@ receivers:
         static_configs:
         # TODO: make this configurable
         - targets: ["localhost:2112"]
+        # Do not relabel job and instance labels if existed.
+        honor_labels: true
         metric_relabel_configs:
           - source_labels: [ __name__ ]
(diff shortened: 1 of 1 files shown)


imjasonh (approved):

mattmoor:
nice :)

PR #70, opened 2024-01-11
Have subscriptions live indefinitely.

mattmoor:
By default pubsub subscriptions live for 31d since their last activity.  We don't want our subscriptions to expire.


mattmoor:
ref: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/pubsub_subscription.html#expiration_policy

tcnghia (approved):

PR #72, opened 2024-01-16
networking: Allow configurable netnum offsets.

wlynch:
Previously this would try to assign a CIDR block starting from 1.
e.g. given cidr = "10.0.0.0/8" and regions = ["a", "b"], you would get:

a: "10.1.0.0/16"
b: "10.2.0.0/16"

Since this always starts at 1, this makes it difficult to be able to select a subnet range. This change adds a field to start subnet numbering at a specific number (starting at 0).

e.g. with the previous example and netnum_offset = 10, you would now get:

a: "10.10.0.0/16"
b: "10.11.0.0/16"

This is a breaking change if downstream consumers are sensitive to subnet netnum needing to start at 1, but can easily be fixed by adding the netnum_offset. I think we want to do this long term so we don't leave IPs with 0 unassigned.

### DIFF
--- a/hack/update-docs.sh
+++ b/hack/update-docs.sh
@@ -8,5 +8,6 @@ for d in `find . -name '*.tf' -exec dirname {} \; | sort | uniq`; do
     terraform-docs markdown table \
         --output-file README.md \
         --output-mode inject \
+        --lockfile=false \
         $d
 done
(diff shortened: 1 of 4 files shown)


mattmoor (approved):

mattmoor:
@wlynch you need to update docs for the new field

PR #84, opened 2024-01-25
rgs: Add missing permissions for Cloud Run to connect to shared VPCs.

wlynch:
- Added roles/compute.networkUser for Cloud Run Service Agent (owned by Cloud Run itself)
- Added roles/compute.networkViewer for service account.

 See https://cloud.google.com/run/docs/configuring/shared-vpc-direct-vpc#set_up_iam_permissions

### DIFF
--- a/modules/regional-go-service/network-iam.tf
+++ b/modules/regional-go-service/network-iam.tf
@@ -1,0 +1,37 @@
+// This file gramnts permissions for Cloud Run to access VPCs.
+// In particular, this is configured to support connecting to a shared VPC in another project,
+// but should be safe to apply in the same project.
+// See https://cloud.google.com/run/docs/configuring/shared-vpc-direct-vpc#set_up_iam_permissions for more details.
+
+data "google_project" "project" {
+  project_id = var.project_id
+}
+
+// Grant Cloud Run the service access to the network.
+resource "google_project_iam_member" "cloudrun_service_network_user" {
+  project = var.network_project
+  member  = "serviceAccount:service-${data.google_project.project.number}@serverless-robot-prod.iam.gserviceaccount.com"
+  role    = "roles/compute.networkUser"
+}
+
+// Grant service account access to see networks on the network project.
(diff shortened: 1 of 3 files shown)


imjasonh (approved):

modules/regional-go-service/network-iam.tf
```diff
@@ -0,0 +1,37 @@
...
+data "google_project" "project" {
+  project_id = var.project_id
+}
+
+// Grant Cloud Run the service access to the network.
+resource "google_project_iam_member" "cloudrun_service_network_user" {
+  project = var.network_project
+  member  = "serviceAccount:service-${data.google_project.project.number}@serverless-robot-prod.iam.gserviceaccount.com"
+  role    = "roles/compute.networkUser"
+}
```
mattmoor, line 15: There are no values of this specific to this invocation of the module, which tells me that this is likely the wrong place for it.

I don't even think `networking` is the right place for it, I suspect that setting this up in `chainguard-dev/infra` to enable Cloud Run to utilize cross-project networking is probably what we want.

modules/regional-go-service/network-iam.tf
```diff
@@ -0,0 +1,37 @@
...
+
+// Grant service account access to see networks on the network project.
+resource "google_project_iam_member" "project_network_viewer" {
+  project = var.network_project
+  member  = "serviceAccount:${var.service_account}"
+  role    = "roles/compute.networkViewer"
+}
+
+// Grant service account access to use subnet. This is typically granted with roles/run.serviceAgent,
```
mattmoor, line 24: I thought I posted this comment already, but I suspect that we are using the wrong service_account below because that isn't granted `run.serviceAgent`, so I suspect that this should be granted to:
```
service-${data.google_project.project.number}@serverless-robot-prod.iam.gserviceaccount.com
```

wlynch:
Closing for https://github.com/chainguard-dev/terraform-infra-common/pull/86

PR #86, opened 2024-01-26
networking: Grant Cloud Run p4sa access to shared network resources.

wlynch:
- Configures Cloud Run service account access to network resources based on https://cloud.google.com/run/docs/configuring/shared-vpc-direct-vpc#set_up_iam_permissions.
- Creates route to private.googleapis.com based on https://cloud.google.com/vpc/docs/configure-private-google-access#config-options
- Adds "egress-inet" label to default internet gateway to require resources to opt-in to external internet egress.


modules/networking/cloudrun.tf
```diff
@@ -0,0 +1,30 @@
...
+resource "google_compute_subnetwork_iam_member" "subnet_network_user" {
+  for_each = resource.google_compute_subnetwork.regional
+
+  // If not set, provider project should be used.
+  project    = var.project_id
+  region     = each.value.region
+  subnetwork = each.value.name
+  role       = "roles/compute.networkUser"
+  member     = "serviceAccount:service-${data.google_project.cloudrun_service_project.number}@serverless-robot-prod.iam.gserviceaccount.com"
+}
```
mattmoor, line 30: I don't think this belongs here.  It feels like we need this in exactly the contexts where we aren't using the networking module?

wlynch: My understanding is the networking module sets up all the resources we need for a VPC network. In that sense, what might make more sense is to move out the network creation out of the module and always require users to pass in the network id so we can configure everything else around it. To me these settings are more in line with the Cloud Run DNS + peering setup, which IIUC is also a one time per-environment configuration we need to make (instead of per-service).

Re breaking changes - IIUC I think this is an additive change because it's using iam_member instead of iam_binding? If the network project matches the service project, then this is granting permissions the Cloud Run service account already has, which is safe to do.
I know you had concerns around if Cloud Run services are removed we also yanking the permissions, but if I'm understanding the intended usage of the network module those won't apply here since there should only be one for each environment?
lmk if there's something I'm missing though 🤔

modules/networking/main.tf
```diff
@@ -1,21 +1,39 @@
 // Create a global network in which to place our resources.
 resource "google_compute_network" "this" {
+  // Only create network if one was not explicitly passed in.
+  count                           = var.network != null ? 0 : 1
```
mattmoor, line 4: this is a breaking change

modules/networking/cloudrun.tf
mattmoor: I commented on the piece that is breaking below.

My point around the multi-project stuff (including this diff) is that it's irrelevant to the single project case this is designed for and we are bolting parts onto this and overloading it's purpose to meet the needs of our strange downstream setup.

Rather than scope creeping this module and trying to call it downstream, I think that we should just be trying to mirror it's functionality within our strange multi-project setup.

PR #87, opened 2024-01-26
some updates

cpanato:
- group dependabot updates
- update golangci-lint
- update go version in go.mod
- update check for public-services-are-unauthenticated resource

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,17 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    groups:
+      gomod:
+        update-types:
+          - "minor"
+          - "patch"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       interval: "daily"
+    groups:
+      actions:
+        update-types:
+          - "minor"
+          - "patch"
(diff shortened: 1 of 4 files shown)


modules/regional-go-service/outputs.tf
```diff
@@ -0,0 +1,4 @@
+output "cloud_run_endpoint" {
+  description = "Cloud Run services endpoint"
+  value       = zipmap(values(google_cloud_run_v2_service.this).*.location, values(google_cloud_run_v2_service.this).*.uri)
+}
```
mattmoor, line 4: While it's annoying, I had specifically avoided exposing this to ~force folks to go through `authorize-private-service`, and I'd kind of like to keep that.

However, you should be able to do something similar to that to look it up once this has deployed:
https://github.com/chainguard-dev/terraform-infra-common/blob/7112a5ef4814be6321fed4ef75f85ba54645bd37/modules/authorize-private-service/main.tf#L10-L16

mattmoor (commented):
Really just one comment, but otherwise LGTM

modules/regional-go-service/outputs.tf
cpanato: got it, i will drop that

cpanato: @mattmoor ptal

mattmoor (approved):

PR #99, opened 2024-02-01
add error reporting widget and add it to service dashboard

k4leung4:
will look something like below

Screenshot 2024-01-31 at 8 30 15 PM

cloud run has its own error reporting grouping, have it show in the dashboard
will cover various errors that cloudrun detects 

usage will need to be updated to pass in the project_id


k4leung4:
alerting will still be log based, as error reporting doesnt support filtering nor terraform
https://github.com/hashicorp/terraform-provider-google/issues/12068

imjasonh (approved):
Nice!

mattmoor (approved):

PR #100, opened 2024-02-01
Change the type of dashboard var.alerts

mattmoor:
Only the key-space / size of a map needs to be known at plan time, so this should be safe to for_each 🤞


k4leung4 (approved):

mattmoor:
rebased to pick up the `var` change in Kenny's PR so I can test this 🤞

mattmoor:
Nice, with this I can overlay multiple probers into a single dashboard: 
image

PR #114, opened 2024-02-03
add error reporting to cloudevent receiver

k4leung4:
matches that of service dashboard


mattmoor (approved):
Has some validation errors

Issue #117, opened 2024-02-05
Configure an alert policy for when the DTS job is failing


mattmoor:
@k4leung4 I think that we should move the alert policy we have for this in mono into `cloudevent-recorder` (scoped to those DTS jobs).  Can you link to where we configure these today and/or shoot over a PR 🤞

mattmoor:
For context, I noticed that the Octo STS DTS job was failing, and I'd have loved an alert about it 🤞

PR #118, opened 2024-02-05
bigquery dts alert, successful run within 30min

k4leung4:
fix: https://github.com/chainguard-dev/terraform-infra-common/issues/117

### DIFF
--- a/modules/cloudevent-recorder/bigquery.tf
+++ b/modules/cloudevent-recorder/bigquery.tf
@@ -99,3 +99,42 @@ resource "google_bigquery_data_transfer_config" "import-job" {
     delete_source_files             = false
   }
 }
+
+// Alert when no successful run in 30min, it should be successful every 15min
+resource "google_monitoring_alert_policy" "bq_dts" {
+  for_each = local.regional-types
+
+  // Close after 7 days
+  alert_strategy {
+    auto_close = "604800s"
+  }
+
+  combiner = "OR"
+
+  conditions {
+    condition_absent {
+      aggregations {
+        alignment_period     = "1800s"
+        cross_series_reducer = "REDUCE_MAX"
+        per_series_aligner   = "ALIGN_MAX"
+      }
+
+      duration = "1800s"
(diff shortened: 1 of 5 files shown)


modules/cloudevent-recorder/bigquery.tf
```diff
@@ -99,3 +99,42 @@ resource "google_bigquery_data_transfer_config" "import-job" {
     delete_source_files             = false
   }
 }
+
+// Alert when no successful run in 30min, it should be successful every 15min
+resource "google_monitoring_alert_policy" "bq_dts" {
+  for_each = local.regional-types
```
mattmoor, line 105: This could end up being a lot of alerts polluting the dashboard.

What if instead we used a `dynamic "conditions"` block below and use a single alert over all of the DTS configs?

modules/dashboard/cloudevent-receiver/variables.tf
```diff
@@ -27,3 +27,9 @@ variable "notification_channels" {
   type        = list(string)
   default     = []
 }
+
+variable "alerts" {
+  description = "A mapping from alerting policy names to the alert ids to add to the dashboard."
+  type        = map(string)
+  default     = {}
+}
```
mattmoor, line 35: We still need to use this the way we do in the `service` dashboard

modules/cloudevent-recorder/bigquery.tf
mattmoor: We could also do something like one per type, with a `dynamic` block over the regions

mattmoor:
This is gonna be awesome, I'm excited 😁

modules/cloudevent-recorder/bigquery.tf
k4leung4: agree, let me try to combine all the regions to a single alert.
would make it clearer in alerts what is failing

mattmoor (approved):

mattmoor:
amazing

PR #120, opened 2024-02-05
cloudevent-trigger: Add RetryPolicy, ExpirationPolicy blocks.

wlynch:
Adds optional blocks for configuring pubsub retries.

### DIFF
--- a/modules/cloudevent-trigger/main.tf
+++ b/modules/cloudevent-trigger/main.tf
@@ -78,7 +78,18 @@ resource "google_pubsub_subscription" "this" {
     }
   }
 
-  expiration_policy {
-    ttl = "" // This does not expire.
+  dynamic "expiration_policy" {
+    for_each = [var.expiration_policy]
+    content {
+      ttl = expiration_policy.value.ttl
+    }
+  }
+
+  dynamic "retry_policy" {
+    for_each = [var.retry_policy]
+    content {
+      minimum_backoff = retry_policy.value.minimum_backoff
+      maximum_backoff = retry_policy.value.maximum_backoff
+    }
   }
 }
--- a/modules/cloudevent-trigger/variables.tf
+++ b/modules/cloudevent-trigger/variables.tf
@@ -23,3 +23,22 @@ variable "private-service" {
     region = string
   })
 }
+
+variable "expiration_policy" {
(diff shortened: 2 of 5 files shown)


tcnghia (approved):

modules/cloudevent-trigger/main.tf
```diff
@@ -78,7 +78,18 @@ resource "google_pubsub_subscription" "this" {
...
+  dynamic "expiration_policy" {
+    for_each = [var.expiration_policy]
+    content {
+      ttl = expiration_policy.value.ttl
+    }
+  }
+
+  dynamic "retry_policy" {
+    for_each = [var.retry_policy]
+    content {
```
mattmoor, line 90: FWIW `for_each` over a single element isn't really useful, this can be simplified.

modules/cloudevent-trigger/main.tf
```diff
@@ -78,7 +78,18 @@ resource "google_pubsub_subscription" "this" {
     }
   }
 
-  expiration_policy {
-    ttl = "" // This does not expire.
+  dynamic "expiration_policy" {
+    for_each = [var.expiration_policy]
+    content {
```
mattmoor, line 83: same here

modules/cloudevent-trigger/variables.tf
```diff
@@ -23,3 +23,22 @@ variable "private-service" {
...
+
+variable "expiration_policy" {
+  description = "The expiration policy for the subscription."
+  type = object({
+    ttl = optional(string, null)
+  })
+  default = {
+    ttl = "" // This does not expire.
+  }
+}
```
mattmoor, line 35: Honestly, I struggle to reason about a use case where we would want the subscription itself to expire, which is what this is.  I'd rather omit this than invite the confusion of folks assuming it's a message expiration setting.

tcnghia: This probably needs to be `message_retention_duration` instead?

mattmoor: I don't think we want that either because AIUI that will apply to messages where delivery has never been attempted.

mattmoor:
With the DLQ stuff that landed last night (and rolled out to staging this morning) the event handlers are much happier.

PR #128, opened 2024-02-07
Add instance to metrics

tcnghia:
Previously I removed this because it didn't seem important to know the instance ID. However, `instance` turns out to be a uniqueness constraint that GMP treats specially.

Adding back to avoid time series collision issue.


tcnghia:
In my dev env, this has reduced the duplicate metrics error significantly 

image

hectorj2f (approved):

cpanato (approved):

mattmoor (approved):

PR #140, opened 2024-02-15
Add a configmap abstraction

mattmoor:


modules/configmap/README.md
```diff
@@ -0,0 +1,102 @@
...
+        mount_path = "/var/run/foo/"
+      }]
+    }
+  }
+  volumes = [{
+    name = "foo"
+    secret = {
+      secret = module.my-configmap.secret_id
+      items = [{
+        version = module.my-configmap.version
```
imjasonh, line 53: Do we want to encourage pinning to a particular version like this?

mattmoor: I'd say there are trade-offs that folks need to think through.  By putting the version here, then changing the secret will trigger a new version, which will trigger a rollout.  By putting "latest" here, users can only pick up new versions if they use volumes and watch for changes.

As we saw today, this can have rollback implications at the service level.  Perhaps for `configmap` we have this use `deletion_policy = "ABANDON"` so that instead of deleting the older version we orphan it, then we can maybe have our cake and eat it too?

mattmoor: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/secret_manager_secret_version#deletion_policy

mattmoor: I like that, I'm going to do it 🤓

PR #141, opened 2024-02-15
Add the project number form to alert on

mattmoor:


mattmoor:
When I tried to trigger the alert in staging, it didn't show up because it appears in the audit logs with the project number form, so check for both.

k4leung4 (approved):

PR #155, opened 2024-02-21
Expand the list of modules we validate.

mattmoor:


.github/workflows/validate.yaml
```diff
@@ -22,6 +23,10 @@ jobs:
           - dashboard/service
           - dashboard/job
           - dashboard/cloudevent-receiver
+          - prober
```
imjasonh, line 26: Can we make this automatic instead of relying on reviewers to spot it missing?

mattmoor: Yup, it's automatic in mono.  Nghia said he's gonna port it here.

PR #162, opened 2024-02-22
This fixes the cloud run policies in a few ways.

mattmoor:
1. Switch `logName` to admin activity (excl. system events and data access)

The `logName` addition pointing at `data_access` excluded mutations, which is the key thing we want to monitor, and most of what that left was things we want to ignore because GetService is pretty benign and common via the console.

2. Split a separate alert policy off for RunJob monitoring

This is the main exception that IS data_access, so for simplicity just make this it's own alert policy.

### DIFF
--- a/modules/cron/main.tf
+++ b/modules/cron/main.tf
@@ -206,7 +206,7 @@ resource "google_monitoring_alert_policy" "anomalous-job-access" {
 
     condition_matched_log {
       filter = <<EOT
-      logName="projects/prod-enforce-fabc/logs/cloudaudit.googleapis.com%2Fdata_access"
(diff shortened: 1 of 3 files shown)


modules/cron/main.tf
```diff
@@ -206,7 +206,7 @@ resource "google_monitoring_alert_policy" "anomalous-job-access" {
 
     condition_matched_log {
       filter = <<EOT
-      logName="projects/prod-enforce-fabc/logs/cloudaudit.googleapis.com%2Fdata_access"
```
mattmoor, line 209: whoops, this had the project id in it too

k4leung4 (approved):

PR #166, opened 2024-02-23
add bucket-events module

### LINKED ISSUE #142: module to publish to broker on GCS events
GCS is able to publish to PubSub. We should have a lil service subscribe to those events and re-publish them (possibly without transformation?) to a cloudevent broker so that regional-go-services can subscribe in the standard way.

Inputs:
- bucket name
- broker name
- regions? Should it just exist in the bucket region?

```
resource "google_storage_notification" "notification" {
  bucket         = data.google_storage_bucket.bucket.name
  payload_format = "JSON_API_V1"
  topic          = google_pubsub_topic.topic.name
  event_types    = ["OBJECT_FINALIZE", "OBJECT_DELETE"]
}
```

imjasonh:
This module sets up a Pub/Sub subscription for GCS events coming from a bucket, and translates them to CloudEvents that it sends to a configured Broker.

So far I've only tested this with a regional GCS bucket and single-region topic and trampoline service. For multi-regional buckets we arbitrarily pick a region in that multi-region that must be part of the regional-networks. This isn't perfect but it's a first draft.

Fixes #142

### DIFF
--- a/modules/bucket-events/README.md
+++ b/modules/bucket-events/README.md
@@ -63,6 +63,8 @@ No requirements.
 |------|--------|---------|
 | <a name="module_audit-serviceaccount"></a> [audit-serviceaccount](#module\_audit-serviceaccount) | ../audit-serviceaccount | n/a |
 | <a name="module_authorize-delivery"></a> [authorize-delivery](#module\_authorize-delivery) | ../authorize-private-service | n/a |
+| <a name="module_this"></a> [this](#module\_this) | ../regional-go-service | n/a |
+| <a name="module_trampoline-emits-events"></a> [trampoline-emits-events](#module\_trampoline-emits-events) | ../authorize-private-service | n/a |
 
 ## Resources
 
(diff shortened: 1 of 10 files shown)


modules/bucket-events/README.md
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// Listen to events.
+module "bucket-events" {
+  source = "chainguard-dev/common/infra//modules/bucket-events"
+
+  name       = "my-broker"
+  project_id = var.project_id
+
+  bucket   = data.google_storage_bucket.bucket.name
```
mattmoor, line 37: If you just pass this in, then the module can use the `data` source and you can simplify the interface (drop location)

modules/bucket-events/main.tf
```diff
@@ -0,0 +1,214 @@
...
+  message_retention_duration = "600s"
+
+  message_storage_policy {
+    allowed_persistence_regions = [each.key]
+  }
+}
+
+// Create a notification to the topic.
+resource "google_storage_notification" "notification" {
+  for_each       = var.regions
```
mattmoor, line 195: You don't want this.  It will mean you get N of the event instead of 1.  I think you want the single region colocated with the bucket.

modules/bucket-events/main.tf
```diff
@@ -0,0 +1,214 @@
+terraform {
+  required_providers {
+    ko     = { source = "ko-build/ko" }
+    cosign = { source = "chainguard-dev/cosign" }
+  }
+}
+
```
mattmoor, line 7: delete

modules/bucket-events/main.tf
imjasonh: Okay that makes sense. So you'd expect this module to take a us-central1 bucket and create a us-central1 topic/subscription/service that publishes to an ingress also only in that region?

I had started more along those lines, but thought we'd want to unshackle ourselves from regions as soon as possible.

But I can go back that way too. For multi-region buckets (`US`, `EU`) do we want multi-regional subscriber setups?

modules/bucket-events/main.tf
```diff
@@ -0,0 +1,214 @@
...
+  length  = 4
+  upper   = false
+  special = false
+}
+
+// A dedicated service account for this subscription.
+resource "google_service_account" "this" {
+  project = var.project_id
+
+  account_id   = "${var.name}-${random_string.suffix.result}"
```
mattmoor, line 18: I think that maybe this (and below) should just be `var.name` and the burden of unique naming should be on the caller.  WDYT?

modules/bucket-events/main.tf
```diff
@@ -0,0 +1,214 @@
...
+
+module "this" {
+  source     = "../regional-go-service"
+  project_id = var.project_id
+  name       = "${var.name}-trampoline"
+  regions    = var.regions
+
+  service_account = google_service_account.this.email
+  containers = {
+    "recorder" = {
```
mattmoor, line 74: `recorder` -> `adapter`?

modules/bucket-events/main.tf
mattmoor: I think that's what we want.  It makes data sovereignty considerations a lot simpler and practically speaking with geographic load balancing it'd hit the colocated service anyways, or (like this) we will do Nx the amount of work because EVERY region will get a subscription event instead of just the one the bucket lives in.

re: multi-regional buckets

With egress cost becoming a thing for muilti-regional access within that multi-region, I've been gravitating towards regional storage for basically everything.

modules/bucket-events/variables.tf
```diff
@@ -0,0 +1,44 @@
...
+  type = string
+}
+
+variable "bucket" {
+  description = "The name of the bucket to watch for events."
+  type        = string
+}
+
+variable "regions" {
+  description = "A map from region names to a network and subnetwork.  A pub/sub topic and ingress service (publishing to the respective topic) will be created in each region, with the ingress service configured to egress all traffic via the specified subnetwork."
```
mattmoor, line 15: update description?  looks like copy/pasta

modules/bucket-events/main.tf
```diff
@@ -0,0 +1,213 @@
...
+  source = "../authorize-private-service"
+
+  project_id = var.project_id
+  region     = local.region
+  name       = var.name
+
+  service-account = google_service_account.delivery.email
+}
+
+resource "google_pubsub_topic" "dead-letter" {
```
mattmoor, line 109: I think the `cloudevent-trigger` surfaces this as an output

modules/bucket-events/cmd/trampoline/main.go
```diff
@@ -0,0 +1,118 @@
...
+
+	log.Infof("env: %+v", env)
+
+	c, err := idtoken.NewClient(ctx, env.IngressURI)
+	if err != nil {
+		log.Fatalf("failed to create idtoken client: %v", err) //nolint:gocritic
+	}
+	ceclient, err := cloudevents.NewClientHTTP(
+		cloudevents.WithTarget(env.IngressURI),
+		cehttp.WithRoundTripper(httpmetrics.WrapTransport(c.Transport)))
```
mattmoor, line 59: Can you use `cehttp.WithClient` instead?

mattmoor (approved):
disabled auto-merge for comments, but since they are all superficial I am approving for when those are done. 🙏

modules/bucket-events/main.tf
imjasonh: I don't get to use a trigger though, I'm dealing with a raw PubSub topic and a GCS notification. Let me know if I'm missing something though.

mattmoor: I was just saying that it might make sense to output it in a similar way here

PR #173, opened 2024-02-28
cloudevent-trigger: support filter prefix

imjasonh:
https://cloud.google.com/pubsub/docs/subscription-message-filter#filtering_syntax

This adds a new input var, `filter_prefix`, which works like `filter` but with attribute prefixes instead of exact matches.


imjasonh:
Another approach would be to expose the direct Pub/Sub filter and concatenate `filter` with `raw_filter` or whatever (or make them mutually exclusive).

This is a bit similar to Knative's new beta filtering scheme https://knative.dev/docs/eventing/experimental-features/new-trigger-filters/, which is based on a cloudevents spec: https://github.com/cloudevents/spec/blob/main/subscriptions/spec.md#324-filters -- instead of CE's dialects though we'd just expose direct Pub/Sub filtering as documented by GCP

mattmoor (approved):

PR #199, opened 2024-03-11
github-events: forward GitHub webhook events to broker

imjasonh:


modules/github-events/cmd/trampoline/main.go
```diff
@@ -0,0 +1,113 @@
...
+		defer r.Body.Close()
+
+		// https://docs.github.com/en/webhooks/using-webhooks/validating-webhook-deliveries
+		payload, err := github.ValidatePayload(r, []byte(env.WebhookSecret))
+		if err != nil {
+			log.Errorf("failed to verify webhook: %v", err)
+			w.WriteHeader(http.StatusForbidden)
+			fmt.Fprintf(w, "failed to verify webhook: %v", err)
+			return
+		}
```
imjasonh, line 67: @wlynch This is failing for me, and I'd love some help debugging why. I've confirmed the rest of it works as expected, but I might be doing something stupid with the webhook secret.

modules/github-events/README.md
```diff
@@ -0,0 +1,102 @@
...
+  name       = "github-events"
+  regions    = module.networking.regional-networks
+  ingress    = module.cloudevent-broker.ingress
+
+  // Which user is allowed to populate webhook secret values.
+  secret_version_adder = "user:you@company.biz"
+}
+```
+
+After applying this, generate a random secret value and add it to the GitHub
```
joshrwolf, line 53: can we use github tf provider to create the webhook and populate the secret?

found-it: Something like this https://github.com/chainguard-dev/infra-images/pull/648/files#diff-8789f1c464f781d4f07e97ec136d38681e1b4c20108a5eaaa5fbfe7f110efbfe

imjasonh: Yeah, that could be done by the caller of this module. Maybe the module should take an input of the webhook secret, which it could populate the initial version of the secret with.

I'm a little uneasy about TF doing secret rotation, since it would need to coordinate between both places and any time in between we'd drop messages.

modules/github-events/main.tf
```diff
@@ -0,0 +1,80 @@
...
+        working_dir = path.module
+        importpath  = "./cmd/trampoline"
+      }
+      ports = [{ container_port = 8080 }]
+      env = [{
+        name = "WEBHOOK_SECRET"
+        value_from = {
+          secret_key_ref = {
+            name    = module.webhook-secret.secret_id
+            version = "latest"
```
found-it, line 56:
```suggestion
        value_source = {
          secret_key_ref = {
            secret  = module.webhook-secret.secret_id
            version = "latest"
```

found-it: This is why the validation is failing. The key is populated into the environment but with an empty value.

modules/github-events/cmd/trampoline/main.go
found-it: [See my comment below](https://github.com/chainguard-dev/terraform-infra-common/pull/199#discussion_r1520525203)

modules/github-events/cmd/trampoline/main.go
```diff
@@ -0,0 +1,113 @@
...
+
+		// https://docs.github.com/en/webhooks/webhook-events-and-payloads#delivery-headers
+		t := r.Header.Get("X-GitHub-Event")
```
found-it, line 70: You should be able to use `github.WebHookType(r)` as well

modules/github-events/cmd/trampoline/main.go
```diff
@@ -0,0 +1,113 @@
...
+		if t == "" {
+			log.Errorf("missing X-GitHub-Event header")
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+		t = "dev.chainguard.github." + r.Header.Get("X-GitHub-Event")
```
found-it, line 76: There is a `ping` event type.. do we want to capture that or just ack it and keep it from cluttering the event bus?

Similar to [this](https://github.com/chainguard-dev/infra-images/blob/07c6db8b5f57dfcd174c451249aa44e96d9615c1/quality/be-better/main.go#L49-L51)

modules/github-events/cmd/trampoline/main.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+
+		// https://docs.github.com/en/webhooks/webhook-events-and-payloads#delivery-headers
+		t := github.WebHookType(r)
+		if t == "" {
+			log.Errorf("missing X-GitHub-Event header")
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+		t = "dev.chainguard.github." + r.Header.Get("X-GitHub-Event")
```
found-it, line 76:
```suggestion
		t = "dev.chainguard.github." + t
```

found-it: minor nit, in case the event header ever inexplicably changes from the WebHookType(r)

found-it (approved):

modules/github-events/main.tf
```diff
@@ -0,0 +1,90 @@
...
+  region     = each.key
+  name       = var.ingress.name
+
+  service-account = google_service_account.service.email
+}
+
+data "google_cloud_run_v2_service" "this" {
+  for_each = var.service-ingress == "INGRESS_TRAFFIC_ALL" ? var.regions : {}
+  project  = var.project_id
+  location = each.key
```
found-it, line 77:
```suggestion
```

PR #200, opened 2024-03-12
[cloudevent-recorder] Add partition_field option.

wlynch:
Allows callers to pass in an optional partition field to instruct BQ for how to partition the data. If omitted, default behavior is used (import time).

### DIFF
--- a/modules/cloudevent-recorder/variables.tf
+++ b/modules/cloudevent-recorder/variables.tf
@@ -52,5 +52,6 @@ variable "types" {
     schema                = string
     alert_threshold       = optional(number, 50000)
     notification_channels = optional(list(string), [])
+    partition_field       = optional(string)
   }))
 }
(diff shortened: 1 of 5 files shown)


modules/cloudevent-recorder/variables.tf
```diff
@@ -52,5 +52,6 @@ variable "types" {
     schema                = string
     alert_threshold       = optional(number, 50000)
     notification_channels = optional(list(string), [])
+    partition_field       = optional(string)
```
k4leung4, line 55:
```suggestion
    partition_field       = optional(string, "")
```

wlynch: I don't think we want this - we want to default to null if the field isn't set so we keep compatibility to the current behavior.

Setting "" may cause the GCP API to treat this as a set value and return an error.

If you want I can make this `optional(string, null)` to be explicit?

k4leung4 (approved):

PR #213, opened 2024-03-19
Change our attribute mapping to fit in character limits.

mattmoor:
The `google.subject` claim is limited to 127 characters, which for certain repo names (e.g. `chainguard-dev/terraform-provider-chainguard`) is too small.

To mitigate this, we dump the repository name being explicitly separate and instead rely on the name that prefixes the `workflow_ref`.

There are a handful of additional changes here:
1. Switch to `[|]` for escaping the pipes, so they are not treated as disjunctions,
2. Wrap the inner regex in parenthesis, so `^(...)$` matches the hole string and the `|` characters don't mistakenly become disjunctions,
3. Include the (missing!) repository prefix in our GSA audit regular expression.

I tested this by modifying the existing secret alert policies to use this and it properly filters the authorized usage:
```
protoPayload.authenticationInfo.principalSubject=~"^(principal://iam\.googleapis\.com/projects/12758742386/locations/global/workloadIdentityPools/github-pool/subject/chainguard-dev/enterprise-packages[|]refs/heads/main[|]chainguard-dev/enterprise-packages/\.github/workflows/(build|withdraw-packages)\.yaml)$"
```

note: the above isn't what we expect the result of this PR to be, but it tests the new escaping, the parenthesis, and the addition of the repo to the workflow prefix.

### DIFF
--- a/modules/github-gsa/main.tf
+++ b/modules/github-gsa/main.tf
@@ -25,7 +25,7 @@ locals {
     )
   )
   # Build up the "workflow" portion of the principal match regular expression.
-  workflowPart = (var.workflow_ref == "*") ? (
+  workflowPart = "${var.repository}/${(var.workflow_ref == "*") ? (
     (var.audit_workflow_ref != "") ? (
       var.audit_workflow_ref
       ) : (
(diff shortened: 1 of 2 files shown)


mattmoor:
Once this lands in a release, we will need to stage the dependabot bump in `chainguard-dev/secrets` so that this doesn't trigger alerts (for too long)

k4leung4 (approved):

PR #215, opened 2024-03-20
feature: optional creation of dummy secret version to avoid bad reference on first deploy

k4leung4:
defaults to false

### DIFF
--- a/modules/secret/main.tf
+++ b/modules/secret/main.tf
@@ -9,6 +9,14 @@ resource "google_secret_manager_secret" "this" {
   }
 }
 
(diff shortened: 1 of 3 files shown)


modules/secret/main.tf
```diff
@@ -9,6 +9,14 @@ resource "google_secret_manager_secret" "this" {
   }
 }
 
+// Create a dummy GCP secret version to avoid bad reference on first deploy.
+resource "google_secret_manager_secret_version" "dummy" {
```
mattmoor, line 13: You need lifecycle annotations or this will reconcile things away

k4leung4: added

PR #216, opened 2024-03-21
add option to declare invokers for cloud run jobs

k4leung4:
also add output of names of services/jobs.

### DIFF
--- a/modules/cron/main.tf
+++ b/modules/cron/main.tf
@@ -46,6 +46,16 @@ resource "ko_build" "image" {
   repo        = local.repo
 }
 
+resource "google_cloud_run_v2_job_iam_member" "invoker" {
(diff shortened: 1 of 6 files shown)


modules/cron/main.tf
```diff
@@ -46,6 +46,16 @@ resource "ko_build" "image" {
   repo        = local.repo
 }
 
+resource "google_cloud_run_v2_job_iam_member" "invoker" {
```
imjasonh, line 49: Should this also exempt the invoker from the [laser grid](https://github.com/chainguard-dev/terraform-infra-common/blob/main/modules/cron/main.tf#L192)?

k4leung4: good point, added exclusion

tcnghia (approved):

PR #226, opened 2024-03-24
feature: option to trigger cloud run cron job on modify

k4leung4:


modules/cron/main.tf
```diff
@@ -6,6 +6,9 @@ terraform {
     google = {
       source = "hashicorp/google"
     }
+    terracurl = {
+      source = "devops-rob/terracurl"
+    }
```
mattmoor, line 11: I'd really like to avoid adding this dependency.

k4leung4: ack

k4leung4: removed the terracurl and replaced it with a null_resource that calls out to gcloud

considering our use of gcp, for dev, one need to auth with gcloud to run, and gcloud is also available via github workflow, so the dependency on gcloud should not be an issue.

does this seem a viable alternative?

tcnghia (approved):

PR #227, opened 2024-03-25
feature: add request failure rate to dashboard

cmdpdx:
- Add a new widget types `xy-ratio` for comparing two time series queries, and `percent` as a wrapper around `xy-ratio` for the common case where the denominator is a subset of the numerator metrics.
- Use `percent` widget for `Request failure rate` widget in HTTP section to measure % responses that are `5xx` for HTTP services.


modules/dashboard/widgets/xy-ratio/main.tf
```diff
@@ -0,0 +1,57 @@
...
+        timeSeriesQuery = {
+          // https://cloud.google.com/monitoring/api/ref_v3/rest/v1/projects.dashboards#timeseriesfilterratio
+          timeSeriesFilterRatio = {
+            numerator = {
+              filter = join("\n", var.numerator_filter)
+              aggregation = {
+                alignmentPeriod    = var.alignment_period
+                perSeriesAligner   = var.numerator_align
+                crossSeriesReducer = var.numerator_reduce
+                groupByFields      = var.numerator_group_by_fields
```
k4leung4, line 35: struggling between letting users choose their own poison [full control], or handle the common case where numerator and denominator has be configured so it aligns and can be divided properly, using the same metric, and just filtered differently

cmdpdx: yeah I feel the same way. On one hand, a ratio could really be any two metrics, but in reality how often will they not have the same baseline.

What about both? I could add a second widget that makes the assumption the numerator is just a more filtered version of the denominator metric to simplify the common case, but leave `xy-ratio` as is in case someone wants full control.

k4leung4 (approved):
thanks, this looks good

PR #238, opened 2024-04-02
log rotate: add missing check for errors during bufio scan and increase buffer size to handle large events on a single line

rawlingsj:
With this we now find previously hidden errors duriung the file scan, here's an example of an error

```
Failed to run the uploader: failed to upload file to blobstore: dev.scan-results.rumble.summary/1712085409975835500, bufio scan error: bufio.Scanner: token too long
```


k4leung4 (commented):
please add unit test

k4leung4 (approved):

imjasonh (approved):

PR #239, opened 2024-04-02
feature: option to boost cpu during startup

tcnghia:


k4leung4 (approved):

modules/regional-go-service/variables.tf
```diff
@@ -64,7 +64,8 @@ variable "containers" {
               memory = string
             }
           ), null)
-          cpu_idle = optional(bool, true)
+          cpu_idle          = optional(bool, true)
+          startup_cpu_boost = optional(bool, false)
```
tcnghia, line 68:
```suggestion
          startup_cpu_boost = optional(bool, true)
```

PR #245, opened 2024-04-05
cloudevent recorder: optionally allow users to specify existing datas…

rawlingsj:
…et and tables to record events into

This helps folks migrate to common infra modules who have existing datasets.

The existing behaviour are the defaults, i.e. without enabling the new create dataset and create table configs nothing will change.

### DIFF
--- a/modules/cloudevent-recorder/recorder.tf
+++ b/modules/cloudevent-recorder/recorder.tf
@@ -96,6 +96,13 @@ module "triggers" {
   notification_channels = var.notification_channels
 }
 
+locals {
+  alerts = tomap({
+    for type, schema in var.types :
+    "BQ DTS ${var.name}-${type}" => lookup(google_monitoring_alert_policy.bq_dts, type, null) != null ? google_monitoring_alert_policy.bq_dts[type].id : null
+  })
+}
+
 module "recorder-dashboard" {
   source       = "../dashboard/cloudevent-receiver"
   service_name = var.name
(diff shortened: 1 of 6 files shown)


modules/cloudevent-recorder/recorder.tf
```diff
@@ -95,6 +95,14 @@ module "triggers" {
   notification_channels = var.notification_channels
 }
 
+locals {
+  alerts = tomap({
+    for type, schema in var.types :
+    "BQ DTS ${var.name}-${type}" => google_monitoring_alert_policy.bq_dts[type].id
```
imjasonh, line 101: Is the idea that this is a temporary step to allow folks to `terraform import` the resource and get back into the normal state?

It'd be nice if we could get the alerts even on existing datasets.

rawlingsj: yeah, I've just made a change, sorry muscle memory amended the previous commit and pushed so it's not obvious my change but if you look at the 
```
locals {
  alerts = tomap({....
```
you can see the change, is that what you're thinking?

tcnghia (approved):

modules/cloudevent-recorder/recorder.tf
mattmoor: I'd love for the goal-state to be what we have today, and not need to keep this bridge logic around ~forever.

Can we import the old data into a new dataset?  Can we switch consumers to use a `UNION` query for a bit?

rawlingsj: There are other defaults like dataset and table expiration that he current tables are created with by default.  The usecase we have for image scans means we'd want to not expire that data.  So being able to provide as existing dataset and tables means folks can set these up with whatever config they want and still benefit from these modules.

We could change downstream queries to handle a `UNION`, or transfer data but that increases the bar to migrate to common infra, which has taken some time already.  If it's an issue then yeah we can do it and will look into that next week, just a lot is depending on a migration asap really.

rawlingsj:
I'm going to close this and leave migrate our queries and dashboards to the new default tables created by the recorder.

PR #246, opened 2024-04-05
cloudevent recorder, add configurable method of syncing events from a…

rawlingsj:
… broker

tl;dr specify a `method` tf var to override the default approach used to write broker events to a GCS bucket.

```
module "foo-emits-events" {
  source = "chainguard-dev/common/infra//modules/cloudevent-recorder"
  method = "gcs"

  ....
}
```

This keeps the default `method` of using cloudevent trigger and  regional go service + log rotate to conksume events from a broker and publish to a GCS bucket.

We add a configurable `method` that allows a user to select `gcs` today, this takes advantage of GCP native approach for a subscriber to write events directly into a GCS bucket, removing the need to run the cloud event trigger services above.  Docs on the [native subscription integration](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/pubsub_subscription#example-usage---pubsub-subscription-push-cloudstorage).

Future additional method of `bq` can be explored which uses a [similar native integration](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/pubsub_subscription#example-usage---pubsub-subscription-push-bq-table-schema).

### DIFF
--- a/modules/cloudevent-recorder/recorder.tf
+++ b/modules/cloudevent-recorder/recorder.tf
@@ -13,7 +13,7 @@ resource "google_service_account" "recorder" {
 // The recorder service account is the only identity that should be writing
 // to the regional GCS buckets.
 resource "google_storage_bucket_iam_binding" "recorder-writes-to-gcs-buckets" {
-  for_each = var.regions
+  for_each = var.method == "trigger" ? var.regions : {}
 
   bucket  = google_storage_bucket.recorder[each.key].name
   role    = "roles/storage.admin"
@@ -21,6 +21,7 @@ resource "google_storage_bucket_iam_binding" "recorder-writes-to-gcs-buckets" {
 }
 
 module "this" {
+  count      = var.method == "trigger" ? 1 : 0
   source     = "../regional-go-service"
   project_id = var.project_id
   name       = var.name
@@ -77,7 +78,7 @@ resource "random_id" "trigger-suffix" {
 
 // Create a trigger for each region x type that sends events to the recorder service.
 module "triggers" {
-  for_each = local.regional-types
+  for_each = var.method == "trigger" ? local.regional-types : {}
 
   source = "../cloudevent-trigger"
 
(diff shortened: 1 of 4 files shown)


rawlingsj:
I do wonder if using this new tf var of `method` as the indication to override the default behavour is best, any better ways?

rawlingsj:
Going to refactor this to make the gcs integration an implementation detail of the recorder rather than exposing it as a useable module.

rawlingsj:
Screenshot 2024-04-05 at 16 49 23

Tested after refactoring the gcs sync into an implementation detail of the recorder.  It's working nicely!

imjasonh (approved):

k4leung4 (approved):

modules/cloudevent-recorder/recorder.tf
```diff
@@ -13,14 +13,15 @@ resource "google_service_account" "recorder" {
...
-  for_each = var.regions
+  for_each = var.method == "trigger" ? var.regions : {}
 
   bucket  = google_storage_bucket.recorder[each.key].name
   role    = "roles/storage.admin"
   members = ["serviceAccount:${google_service_account.recorder.email}"]
 }
 
 module "this" {
+  count      = var.method == "trigger" ? 1 : 0
```
mattmoor, line 24: This is a breaking change anyways that will trigger a recreation of the resources, so why not just go for the migration?

modules/cloudevent-recorder/subscriber-gcs.tf
```diff
@@ -0,0 +1,83 @@
...
+  ack_deadline_seconds = var.ack_deadline_seconds
+
+  filter = "attributes.ce-type=\"${each.value.type}\""
+
+  cloud_storage_config {
+    bucket          = google_storage_bucket.recorder[each.value.region].name
+    filename_prefix = "${each.value.type}/"
+    max_bytes       = var.cloud_storage_config_max_bytes
+    max_duration    = "${var.cloud_storage_config_max_duration}s"
+  }
```
mattmoor, line 59: Should we simply change `cloudevent-trigger` to support delivery to a bucket instead of changing the recorder?

Better still, if we add support for delivery to BigQuery directly with `bigquery_config`, then can we eliminate our need for GCS entirely?

PR #248, opened 2024-04-05
bots: initial commit

imjasonh:
This is a work in progress prototype demonstrating how we can write an army of cute lil bots to manage our repos for us.

The basic idea is that we can write bots that subscribe to GH events (PR opened, commit pushed, workflow finished, etc.), and those bots can take actions on the subject of the event (the PR, the workflow), as easily as possible.

This adds a GH client wrapper that handles the OctoSTS token request/revoke, and tries to simplify the set of actions the bots are trying to do.

The other concept touched on in this PR is "blocking labels": bots can add labels to block merging (e.g., `blocking/dnm`, `blocking/bincapz`, etc.), and the blocker bot will be responsible for seeing those labels and passing/failing a check run on the PR. That check can be made a required check, so that the existence of the labels blocks merging. We can simulate "ack-merging" by removing the labels, which is recorded in the PR's action log.

The idea is that bots have a way to put blocks in place that are intentionally easy to ack (and audited!), and that humans can manually add back as needed (also audited). All of this in Go, a real non-YAMLbash language.

The GH client wrapper also adds a method to create-or-update a comment "owned" by the bot.


modules/github-bots/sdk/bot.go
```diff
@@ -0,0 +1,83 @@
...
+
+		clog.FromContext(ctx).With("event", event).Debugf("received event")
+
+		defer func() {
+			if err := recover(); err != nil {
+				clog.Errorf("stacktrace from panic: %s", debug.Stack())
+			}
+		}()
+
+		if prb, ok := b.(interface {
```
imjasonh, line 47: This is where I hid all the gross stuff. 🤢 

Because events we receive are `schema.PullRequestEvent`s and not full `*github.PullRequestEvent`s, we "cast" the former to the latter by JSON-round-tripping. This works, but it won't be pretty as we expand to more types. It also means bots can't act on a field that we don't explicitly add to `event_types.go` (as BigQuery types 😒), which also makes it recorded in BQ.

modules/github-bots/sdk/bot.go
```diff
@@ -0,0 +1,83 @@
...
+			OnPullRequest(ctx context.Context, pr *github.PullRequest) error
+		}); ok && event.Type() == "dev.chainguard.github.pull_request" {
+			log := clog.FromContext(ctx).With("bot", b.Name(), "event", event.Type())
```
vaikas, line 50: Is this the right place to start collecting metrics on bot response codes, latency, etc.?

imjasonh: This is already collected for us by GCP! We can make dashboards for it, per-bot, or one dashboard for all bots.

vaikas: Ah, fantastic! I just couldn't find any dashboards that had anything besides the high level pubsub which is not that useful from the eventing side since you can't do things like, per event type, per bot, etc.

vaikas: If you can give me a pointer to something where they are, or what to look for, I could take a look at doing a dashboard tmw.

luhring (approved):

imjasonh:
Merging so we can start bringing this up in a general sense in staging.

PR #254, opened 2024-04-08
regional-go-service: expose startup-probe config

rawlingsj:
The default go regional Cloud Run service uses 
```
        startupProbe:
          timeoutSeconds: 240
          periodSeconds: 240
          failureThreshold: 1
          tcpSocket:
            port: 8080
```
I must admit I'm not sure where the 240 was coming from but either way I've found exposing all the config optional useful when starting up services and so they are ready to receive traffic.

I.e. using config like...
```
      startup_probe = {
        initial_delay_seconds = 15
        timeout_seconds       = 2
        period_seconds        = 3
        failure_threshold     = 3
        tcp_socket_port       = 8080
      }
```

Which seems to have really helped reduce start up errors in Cloud Run logs.


k4leung4:
the 240 is the max value for start

```
The startup_probe block supports:

[initial_delay_seconds](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_run_v2_service#initial_delay_seconds) - (Optional) Number of seconds after the container has started before the probe is initiated. Defaults to 0 seconds. Minimum value is 0. Maximum value for liveness probe is 3600. Maximum value for startup probe is 240.
```

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_run_v2_service#initial_delay_seconds

PR #256, opened 2024-04-10
add optional SLO alert for probers and dashboard to track

cmdpdx:
As part of finishing the factory, add an optional alert to probers that triggers when the percent uptime of a service (as measured by the minimum number of successful probes from all regions over a rolling window) drops below the configured SLO threshold.

Also add a dashboard designed to just display alert tiles for all SLO alerts as a centralized view of current SLO status.

### DIFF
--- a/modules/prober/alert.tf
+++ b/modules/prober/alert.tf
@@ -81,7 +81,7 @@ resource "google_monitoring_alert_policy" "slo_alert" {
       filter     = <<-EOT
         metric.type="monitoring.googleapis.com/uptime_check/check_passed"
         resource.type="uptime_url"
-        metric.label.check_id="${local.uptime_check_id}"
+        metric.label.check_id=starts_with("${local.uptime_check_name}")
       EOT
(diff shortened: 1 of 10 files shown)


modules/prober/alert.tf
```diff
@@ -51,3 +51,54 @@ resource "google_monitoring_alert_policy" "uptime_alert" {
...
+    }
+
+    display_name = "${local.uptime_check_name} uptime less than ${local.slo_threshold_friendly}"
+  }
+
+  display_name = "${local.uptime_check_name} uptime less than ${local.slo_threshold_friendly} alert"
+  enabled      = true
+
+  documentation {
+    content = "${local.uptime_check_name} has fallen below ${local.slo_threshold_friendly} over the past day"
```
k4leung4, line 100: a link to slo policy

cmdpdx: Added an optional var to include a link in the documentation.

PR #263, opened 2024-04-11
optionally include GitHub metrics in dashboards

imjasonh:
This also makes it possible for callers to omit the HTTP or GRPC sections, e.g., if the service doesn't serve those.

The default behavior should match what we have today, but dashboards that would benefit from seeing GitHub API metrics can now have that.

GitHub API metrics aren't service-specific, so this graph might be misleading in some cases. At least showing it to people should make it easier to iterate on it.


k4leung4 (approved):
i think this make sense
this allows more customization of the dashboard in a manageable fashion.

PR #268, opened 2024-04-16
Add GitHub App token signer for GCP KMS.

wlynch:
cc @found-it @imjasonh - this is what we're using in a few other places to use kms with github app auth.


jml:

PR #269, opened 2024-04-16
cloudevent-recorder: ignore unknown values

imjasonh:
From this page: https://cloud.google.com/bigquery/docs/cloud-storage-transfer#bq

Screenshot 2024-04-16 at 1 20 11 PM

If this works as documented, we can avoid pre-filtering GH event payloads before they enter the broker, and rely on DTS to drop fields outside our schema all the way "to the right".

We could then make our BQ schemas more of an implementation detail, and internal to schemagen, and GH event subscribers could get the full GH event payload whether or not the fields end up getting recorded.


k4leung4:
im paranoid of a not knowing what is ignored?
if there is a configuration change on events that might cause events to be dropped without us knowing

imjasonh:
> im paranoid of not knowing what is ignored? if there is a configuration change on events that might cause events to be dropped without us knowing

Talked offline, this is a reasonable concern. I've made it a var (default false) so we can only enable it where it's useful, in the GH event recorder

k4leung4 (approved):
thanks

PR #277, opened 2024-04-16
Add the GSA caller's email as a metric dimension.

mattmoor:


tcnghia (approved):

mattmoor:
We should consider having the Ingress add `ce-actor` to the events it admits as well.

PR #284, opened 2024-04-18
Disallow internet egress by default & reroute accounts.google.com -> private.googleapis.com

tcnghia:
* Disallow 0.0.0.0 egress
* Reroute accounts.google.com -> private.googleapis.com so that we can verify OIDC token of Cloud Run private service requests

### DIFF
--- a/modules/networking/dns.tf
+++ b/modules/networking/dns.tf
@@ -34,6 +34,35 @@ resource "google_dns_record_set" "cloud-run-cname" {
 }
 
 // Create a special DNS zone attached to the network in which
+// we will operate our services that reroutes *.google.com
+// that we control.
+resource "google_dns_managed_zone" "private-google-com" {
+  project     = var.project_id
+  name        = "private-google-com-${random_string.suffix.result}"
+  dns_name    = "google.com."
+  description = "This reroutes google.com requests to private.googleapis.com"
+
+  visibility = "private"
+
+  private_visibility_config {
+    networks {
+      network_url = google_compute_network.this.id
+    }
+  }
+}
+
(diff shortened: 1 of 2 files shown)


modules/networking/main.tf
```diff
@@ -7,12 +7,13 @@ resource "google_compute_network" "this" {
   delete_default_routes_on_create = true
 }
 
-// Create a default route to the Internet.
-resource "google_compute_route" "egress-inet" {
+// Allow private Google access from the VPC.
+resource "google_compute_route" "private-google-access" {
```
mattmoor, line 11: Add a `moved` for this to avoid the recreation?

tcnghia: Done

modules/networking/dns.tf
```diff
@@ -33,6 +33,35 @@ resource "google_dns_record_set" "cloud-run-cname" {
...
+  visibility = "private"
+
+  private_visibility_config {
+    networks {
+      network_url = google_compute_network.this.id
+    }
+  }
+}
+
+// Create a record for *.run.app that points to private.googleapis.com
```
tcnghia, line 54:
```suggestion
// Create a record for *.google.com that points to private.googleapis.com
```

mattmoor:
@tcnghia what's left here?

tcnghia:
> @tcnghia what's left here?

Nothing, this just slipped past my mind. Merging now.

tcnghia:
Actually, since this has been tested very long ago I'd run some more testing this week before merging. Don't wait for this PR to cut your release

PR #287, opened 2024-04-19
github-bots: breaking change: replace `source_code` with an embedded …

rawlingsj:
…`containers` variable

github-bots are relatively new.  After using them for a few usecases we required extra env vars and secrets for the bots.  This change allows users to embed a container config that can be used to tailor the bot cloud run service.

### DIFF
--- a/modules/github-bots/main.tf
+++ b/modules/github-bots/main.tf
@@ -21,12 +21,55 @@ variable "broker" {
   type        = map(string)
 }
 
-variable "source_code" {
-  description = "The source code for the bot."
-  type = object({
-    working_dir = string
-    importpath  = string
-  })
+variable "containers" {
+  description = "The containers to run in the service.  Each container will be run in each region."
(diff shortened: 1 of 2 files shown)


modules/github-bots/main.tf
```diff
@@ -21,12 +21,55 @@ variable "broker" {
   type        = map(string)
 }
 
-variable "source_code" {
-  description = "The source code for the bot."
-  type = object({
-    working_dir = string
-    importpath  = string
-  })
+variable "containers" {
```
imjasonh, line 24: I would love to be able to share this structure somehow... 😭

imjasonh (approved):

modules/github-bots/main.tf
rawlingsj: yeah - I'll take a look next week and see if we can make this a bit more DRY

PR #289, opened 2024-04-19
github-bots: extra logging when bots return errors and add extra clou…

rawlingsj:
…d event headers to context so they can be used by bot handlers

### DIFF
--- a/modules/github-bots/sdk/bot.go
+++ b/modules/github-bots/sdk/bot.go
@@ -18,6 +18,15 @@ import (
 	"github.com/kelseyhightower/envconfig"
 )
 
+// Define a type for keys used in context to prevent key collisions.
+type contextKey string
+
+// Define constants for the keys to use with context.WithValue.
+const (
+	ContextKeyAttributes contextKey = "ce-attributes"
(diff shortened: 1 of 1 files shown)


modules/github-bots/sdk/bot.go
```diff
@@ -90,6 +99,15 @@ func Serve(b Bot) {
...
 		}()
 
+		// loop over all event headers and add them to the context so they can be used by the handlers
+		for k, v := range event.Context.GetExtensions() {
+			ctx = context.WithValue(ctx, contextKey(k), v)
+		}
+
+		// add existing event attributes to context so they can be used by the handlers
+		ctx = context.WithValue(ctx, ContextKeyAttributes, event.Extensions())
+		ctx = context.WithValue(ctx, ContextKeyType, event.Type())
```
joshrwolf, line 109: can we scooch this down right before the handler `switch` so we're not needlessly processing events that ultimately get ignored?

rawlingsj: yeah good shout...

joshrwolf (approved):

PR #293, opened 2024-04-20
github-bots/sdk: add helper method for getting attributes from the context

rawlingsj:


modules/github-bots/sdk/bot.go
```diff
@@ -18,13 +18,13 @@ import (
 	"github.com/kelseyhightower/envconfig"
 )
 
-// Define a type for keys used in context to prevent key collisions.
-type contextKey string
+// ContextKey define a type for keys used in context to prevent key collisions.
+type ContextKey string
```
imjasonh, line 22: I think we should keep this private, and add a `AttributeFromContext(ctx)` method instead. This lets us change the implementation without potentially breaking folks that depend on it.

rawlingsj: sounds good, something like this?

hectorj2f (approved):

PR #304, opened 2024-04-24
allow cloud profiler to be used

k4leung4:
a secure way to get profiling data while running on gcp
https://cloud.google.com/profiler/docs/profiling-go

if we decide to adopt `cloud profiler` across the board, we may consider ripping out `pprof` completely

we can use `pprof`, but we would have to register the `pprof` endpoints on the application port, as cloudrun only allows a single port.
we can block external traffic to `/debug/pprof` with cloud armor, and only access via the `*.run.app` host

both options would require enabling
pprof would still require
- tunneling
- curl endpoint with identity token

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	chainguard.dev/sdk v0.1.19
 	cloud.google.com/go/bigquery v1.60.0
 	cloud.google.com/go/compute/metadata v0.3.0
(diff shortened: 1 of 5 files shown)


mattmoor:
Fix issues?

k4leung4:
sorry, ptal

PR #323, opened 2024-05-01
grant project cloudrun viewer only to groups

k4leung4:
follow up from https://github.com/chainguard-dev/internal/pull/3773#discussion_r1585827887

### DIFF
--- a/modules/cron/main.tf
+++ b/modules/cron/main.tf
@@ -195,8 +195,9 @@ resource "google_cloud_run_v2_job_iam_binding" "authorize-calls" {
 }
 
 // project iam, as job iam does allow user to actually list the job to access it
+// only grant it to groups, individual should have access otherwise.
 resource "google_project_iam_member" "authorize-list" {
-  for_each = toset(var.invokers)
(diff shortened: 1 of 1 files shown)


tcnghia (approved):

modules/cron/main.tf
```diff
@@ -195,8 +195,9 @@ resource "google_cloud_run_v2_job_iam_binding" "authorize-calls" {
 }
 
 // project iam, as job iam does allow user to actually list the job to access it
+// only grant it to groups, individual should have access otherwise.
 resource "google_project_iam_member" "authorize-list" {
-  for_each = toset(var.invokers)
+  for_each = toset([for i in var.invokers : i if startswith(i, "group:")])
```
mattmoor, line 200: rather than silently ignoring non-groups, use a lifecycle assertion?

k4leung4: the reason i went with silently ignoring non-groups is because, im using `var.invokers` for granting `run.invoker` and only grant `run.viewer` at the project level for groups.

PR #326, opened 2024-05-02
dashboard: add markdown section with Logs Explorer link

imjasonh:
Screenshot 2024-05-02 at 4 50 48 PM


k4leung4 (commented):
are there cases where we wouldnt want the logurl markdown?
im wondering if it makes sense to provide a markdown module with logurl defaulted to avoid duplication

please fix presubmit

imjasonh:
> are there cases where we wouldnt want the logurl markdown? im wondering if it makes sense to provide a markdown module with logurl defaulted to avoid duplication

I can't think of any, AFAIK we have the logs panel in all these dashboards already. We'll probably need to figure out how to pass custom markdown to these dashboards, like for playbooks etc., but that can happen in a future PR.

...Or it can happen in this PR, if we have ideas about how that should work. 😆

imjasonh:
Deferring this for now. If we want it we can resurrect.

PR #328, opened 2024-05-05
Introduce a `regional-service` module

mattmoor:


imjasonh (commented):
Could we rename regional-go-service and have it accept image||source?

I really don't want to keep more things in sync with the containers type (we already need to for cron and bots, and it sucks)

mattmoor:
@imjasonh we can discuss it, but I worry that optionality will add a bunch of validation complexity (for the `oneof` aspect, conditional builds, etc) and become a breeding ground for subtle bugs.

I also suspect that the `containers` types will naturally vary module to module, e.g. something we didn't consider before, but could be useful when passing a literal container image would be `command`.  In the `[Cron]Job` case, we have differences like not needing regional-env and regional-mounts.

The complexity and inability to share are annoying, but I also think that the types are intentionally divergent across these use cases.

imjasonh:
When we have this, what's the benefit of regional-go-service beyond being a wrapper for a single ko_build and this?

I really don't want to have to think of two things every time we add support for something to one or the other.

mattmoor:
I would say that generally you don’t have to think of both because one calls the other.  You reviewed this before I could push that commit from the crap airplane WiFi (I ultimately had to wait until I got to my hotel).

In terms of the value, today it’s N ko builds, N cosign signs, and eliminating the need to declare those “required providers” in virtually every downstream module we have, which adds up fast.  However, this is also a place where we haven’t gone as deep as I’d like either.  We still can’t attest SBOMs with ko, or do any sort of SLSA provenance attestation.  We should also add in a signature verification check on the base image.

mattmoor:
@imjasonh I opened https://github.com/chainguard-dev/terraform-infra-common/issues/331 to track this, but I don't want to block on this because we need something like this to decompose Octo STS's provisioning, so folks can slot in an image from us.

imjasonh (approved):

PR #338, opened 2024-05-08
Add a clone of idtoken.NewClient() that is more compatible with our metrics Transport

tcnghia:
This function is based on https://github.com/googleapis/google-api-go-client/blob/v0.178.0/idtoken/idtoken.go#L46
but skipping the opts validation that can't be performed here because private types.

We can't just use the `option.WithHTTPClient` and call upstream `idtoken.NewClient()` because that code always tries to cast from the `http.DefaultTransport` anyway, no matter if `option.WithHTTPClient()` is used or not ([see](https://github.com/googleapis/google-api-go-client/blob/v0.178.0/idtoken/idtoken.go#L73)).

### DIFF
--- a/pkg/httpmetrics/transport.go
+++ b/pkg/httpmetrics/transport.go
@@ -52,14 +52,30 @@ func SetBucketSuffixes(bs map[string]string) { bucketSuffixes = bs }
 // Transport is an http.RoundTripper that records metrics for each request.
 var Transport = WrapTransport(http.DefaultTransport)
 
+type MetricsTransport struct {
+	http.RoundTripper
+
+	inner http.RoundTripper
+}
+
 // WrapTransport wraps an http.RoundTripper with instrumentation.
 func WrapTransport(t http.RoundTripper) http.RoundTripper {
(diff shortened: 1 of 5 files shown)


pkg/httpmetrics/transport.go
```diff
@@ -52,14 +52,30 @@ func SetBucketSuffixes(bs map[string]string) { bucketSuffixes = bs }
 // Transport is an http.RoundTripper that records metrics for each request.
 var Transport = WrapTransport(http.DefaultTransport)
 
+type MetricsTransport struct {
+	http.RoundTripper
```
imjasonh, line 56: I don't think we need this if we have inner? Or maybe vice versa? I think it's confusing to have both

tcnghia: `http.RoundTripper` is for the `outer`/`this`. 

I think there may be a better way to do this, but my Go-fu isn't as good so would love to hear how to do this better.

imjasonh (approved):

PR #353, opened 2024-05-15
Add paginated artifact list

found-it:
Now that we are generating more artifacts in the release, they won't all fit into the default page size. This just adds a function to search each page for an artifact before retrieving the next page.

### DIFF
--- a/modules/github-bots/sdk/github.go
+++ b/modules/github-bots/sdk/github.go
@@ -4,6 +4,7 @@ import (
 	"archive/zip"
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -97,7 +98,8 @@ func checkRateLimiting(ctx context.Context, githubErr error) (bool, time.Duratio
 
 	// If GitHub returned an error of type RateLimitError, we can attempt to compute the next time to try the request again
 	// by reading its rate limit information
-	if rateLimitError, ok := githubErr.(*github.RateLimitError); ok {
+	var rateLimitError *github.RateLimitError
+	if errors.As(githubErr, &rateLimitError) {
 		isRateLimited = true
 		retryAfter := time.Until(rateLimitError.Rate.Reset.Time)
 		delay = retryAfter
@@ -105,7 +107,8 @@ func checkRateLimiting(ctx context.Context, githubErr error) (bool, time.Duratio
 	}
 
 	// If GitHub returned a Retry-After header, use its value, otherwise use the default
-	if abuseRateLimitError, ok := githubErr.(*github.AbuseRateLimitError); ok {
+	var abuseRateLimitError *github.AbuseRateLimitError
+	if errors.As(githubErr, &abuseRateLimitError) {
 		isRateLimited = true
 		if abuseRateLimitError.RetryAfter != nil {
 			if abuseRateLimitError.RetryAfter.Seconds() > 0 {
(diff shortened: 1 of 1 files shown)


modules/github-bots/sdk/github.go
```diff
@@ -83,6 +85,60 @@ func (c GitHubClient) Close(ctx context.Context) error {
...
+// returned by the API.
+func handleGithubResponse(ctx context.Context, resp *github.Response, err error) error {
+	// if err is rate limit, delay and try again
+	githubErr := github.CheckResponse(resp.Response)
+	if githubErr != nil {
+		rateLimited, delay := checkRateLimiting(ctx, githubErr)
+		if !rateLimited {
+			return err
+		}
+		fmt.Printf("retrying after %v second delay due to rate limiting\n", delay.Seconds())
```
imjasonh, line 133: Use clog here

modules/github-bots/sdk/github.go
```diff
@@ -83,6 +85,60 @@ func (c GitHubClient) Close(ctx context.Context) error {
...
+	isRateLimited := false
+
+	// If GitHub returned an error of type RateLimitError, we can attempt to compute the next time to try the request again
+	// by reading its rate limit information
+	var rateLimitError *github.RateLimitError
+	if errors.As(githubErr, &rateLimitError) {
+		isRateLimited = true
+		retryAfter := time.Until(rateLimitError.Rate.Reset.Time)
+		delay = retryAfter
+		log.Infof("parsed retryAfter %d from GitHub rate limit error's reset time", retryAfter)
```
imjasonh, line 106: This might be spammy, maybe Debugf?

modules/github-bots/sdk/github.go
```diff
@@ -83,6 +85,60 @@ func (c GitHubClient) Close(ctx context.Context) error {
...
+func handleGithubResponse(ctx context.Context, resp *github.Response, err error) error {
+	// if err is rate limit, delay and try again
+	githubErr := github.CheckResponse(resp.Response)
+	if githubErr != nil {
+		rateLimited, delay := checkRateLimiting(ctx, githubErr)
+		if !rateLimited {
+			return err
+		}
+		fmt.Printf("retrying after %v second delay due to rate limiting\n", delay.Seconds())
+		time.Sleep(delay)
```
imjasonh, line 134: This makes me a little nervous tbh.

GH's rate limit quota resets hourly, so if we blow through it all at once then we could have jobs sleeping for 50m for the reset, costing money while they wait to time out and retry. Multiply that by potentially dozens of replicas, and it's not great.

I think we need a different approach to rate limits in the bot world, and we have some ideas.

found-it: Good note @imjasonh

I don't need the rate limiting to complete my cycle item. Want me to just remove it from this PR and we can chat about the best way to do it moving forward?

modules/github-bots/sdk/github.go
found-it: Good catch!

modules/github-bots/sdk/github.go
joshrwolf: +1, lets separate the contentious rate-limiting parts and land this

found-it: separated ✅

imjasonh (approved):

PR #358, opened 2024-05-17
github-bots handle token error

found-it:
When the octo-sts policy was incorrect we ran into two errors.

The first one in `handleGithubResponse` was a cryptic segfault where `resp` was nil so accessing `resp.Response` threw an error

The second was in the `Close()` function we check if `ts.tok` is nil before trying to revoke. Previously it was never nil, it was an pointer to a Token struct with an empty token. This caused us to fail on token revocation

### DIFF
--- a/modules/github-bots/sdk/github.go
+++ b/modules/github-bots/sdk/github.go
@@ -55,7 +55,14 @@ func (ts *tokenSource) Token() (*oauth2.Token, error) {
 		ctx := context.Background()
 		clog.FromContext(ctx).Debugf("getting octosts token for %s/%s - %s", ts.org, ts.repo, ts.policyName)
 		otok, err := octosts.Token(ctx, ts.policyName, ts.org, ts.repo)
-		ts.tok, ts.err = &oauth2.Token{AccessToken: otok}, err
+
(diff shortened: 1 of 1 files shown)


modules/github-bots/sdk/github.go
```diff
@@ -115,21 +122,23 @@ func checkRateLimiting(_ context.Context, githubErr error) (bool, time.Duration)
...
 func handleGithubResponse(ctx context.Context, resp *github.Response, err error) error {
-	githubErr := github.CheckResponse(resp.Response)
-	if githubErr != nil {
-		rateLimited, delay := checkRateLimiting(ctx, githubErr)
-		// if we were not rate limited, return err
-		if !rateLimited {
-			return err
+	// resp may be nil if err is nonempty. However, err may contain a rate limit
+	// error so we have to inspect for rate limiting if resp is non-nil
+	if resp != nil {
```
stormqueen1990, line 127: Does it make sense to have an early return if `resp == nil`, instead of enclosing the whole block in an if statement?

stormqueen1990: I beg to differ on that, but this is a code style nit too

stormqueen1990 (approved):

PR #359, opened 2024-05-20
Create a dashboard per trigger to stay under widget limit

found-it:
Slack context: https://chainguard-dev.slack.com/archives/C02SD39C6BW/p1715107930859599

We hit widget limits when adding more event types to the recorder. This is a pass at modifying the dashboard module to stay below the widget limit by creating a dashboard per trigger.


joshrwolf (approved):
this lgtm, but we probably want to run it by oncall/infra squad as a heads up?

tcnghia (approved):

PR #362, opened 2024-05-22
Add support for disabling routes

mattmoor:


mattmoor:
I used this to turn down a hostname in my dev environment just now, which uncovered at least one issue (in my first draft, it's covered here), so pretty confident this works!

PR #365, opened 2024-05-23
Allow mounting secrets to prober server

tcnghia:


k4leung4 (commented):
i was thinking something like this
https://github.com/chainguard-dev/mono/blob/main/datastore/iac/iam.tf#L143C1-L149C12

you specify a (name, secret id, version) and plumb it through via env

tcnghia:
> i was thinking something like this https://github.com/chainguard-dev/mono/blob/main/datastore/iac/iam.tf#L143C1-L149C12
> 
> you specify a (name, secret id, version) and plumb it through via env

I used the same syntax as `cron` instead (version = `latest`)

k4leung4 (approved):

PR #393, opened 2024-06-13
bot-sdk: add a new event type and handler for workload run events tha…

rawlingsj:
…t are enriched with a log location cloud event attribute

### DIFF
--- a/modules/github-bots/sdk/handlers.go
+++ b/modules/github-bots/sdk/handlers.go
@@ -22,6 +22,12 @@ func (r WorkflowRunHandler) EventType() EventType {
 	return WorkflowRunEvent
 }
 
+type WorkflowRunLogsHandler func(ctx context.Context, wre github.WorkflowRunEvent) error
+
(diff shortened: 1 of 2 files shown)


modules/github-bots/sdk/handlers.go
```diff
@@ -44,4 +50,5 @@ const (
 
 	// LoFo events
 	WorkflowRunArtifactEvent EventType = "dev.chainguard.lofo.workflow_run_artifacts"
+	WorkflowRunWithLogsEvent EventType = "dev.chainguard.lofo.workflow_run_logs"
```
found-it, line 53: Should we drop the `..With..` to keep it consistent with the artifact event?
```suggestion
	WorkflowRunLogsEvent EventType = "dev.chainguard.lofo.workflow_run_logs"
```

rawlingsj: sounds great!

imjasonh (approved):

PR #413, opened 2024-06-21
octo-sts-secret: New terraform module for rotating GitHub tokens

pdeslaur:
This module allows to store an octo-sts token in a Google Cloud Secret in order to:

1. Allow a Google Cloud service to access a GitHub repo (Google Cloud Build, Google Cloud Dataform)
2. (not recommended) Share a cached GitHub secret between services

The main use case is (1) - but it also makes (2) possible if really needed.

### DIFF
--- a/modules/octo-sts-secret/cmd/rotate/main.go
+++ b/modules/octo-sts-secret/cmd/rotate/main.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"syscall"
 
 	secretmanager "cloud.google.com/go/secretmanager/apiv1"
 	"cloud.google.com/go/secretmanager/apiv1/secretmanagerpb"
--- a/modules/octo-sts-secret/cmd/rotate/main.go
+++ b/modules/octo-sts-secret/cmd/rotate/main.go
@@ -34,7 +34,7 @@ func main() {
 		panic(fmt.Errorf("failed to parse env: %w", err))
 	}
 
-	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer cancel()
 
 	c, err := secretmanager.NewClient(ctx)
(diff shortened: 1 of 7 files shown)


modules/octo-sts-secret/cmd/rotate/main.go
```diff
@@ -0,0 +1,89 @@
...
+	GitHubTokenSecret string `envconfig:"GITHUB_TOKEN_SECRET" required:"true"`
+}
+
+func main() {
+	var env envConfig
+	if err := envconfig.Process("", &env); err != nil {
+		panic(fmt.Errorf("failed to parse env: %w", err))
+	}
+
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
```
k4leung4, line 37:
```suggestion
	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
```
see https://github.com/chainguard-dev/mono/pull/18076

modules/octo-sts-secret/main.tf
```diff
@@ -0,0 +1,55 @@
...
+
+  name       = var.name
+  project_id = var.project_id
+  region     = var.region
+
+  invokers = var.invokers
+
+  timeout = "60s" // 1 minute
+  # Run every 30 minutes
+  schedule = "*/30 * * * *"
```
k4leung4, line 40:
```suggestion
  schedule = "17-59/30 * * * *"
```

nit: add an offset

https://stackoverflow.com/questions/12786410/run-cron-job-every-n-minutes-plus-offset

k4leung4 (commented):
whats the monitoring plan here?
as i assume if the rotation fails, a bunch of workflows will start running into auth issues?

pdeslaur:
> whats the monitoring plan here? as i assume if the rotation fails, a bunch of workflows will start running into auth issues?

Good question! I added a monitoring dashboard to the component. This is using the standard `cron` module, so we get the same monitoring capabilities via notifications. Basically, monitoring can be configured with the rotate job or with the downstream service.

k4leung4:
one more question

does it make sense to keep the previously active secret around but disabled?
the motivation is that if some logic in octosts changed and the new secret is not functioning as expected
there is no rollback story, as the previous secret would be deleted at this point.

pdeslaur:
> one more question
> 
> does it make sense to keep the previously active secret around but disabled? the motivation is that if some logic in octosts changed and the new secret is not functioning as expected there is no rollback story, as the previous secret would be deleted at this point.

That's a good idea - I like it. I'll make the change!

PR #437, opened 2024-07-09
Explicitly set the trace sampling to 10%

tcnghia:


pkg/httpmetrics/metrics.go
```diff
@@ -255,6 +255,8 @@ func tracerOptionsGCP(ctx context.Context) []trace.TracerProviderOption {
 	return []trace.TracerProviderOption{
 		trace.WithResource(res),
 		trace.WithSpanProcessor(bsp),
+		// Use 0.1 percent.
```
cmdpdx, line 258:
```suggestion
		// Use 10 percent.
```
nit

pkg/httpmetrics/metrics.go
```diff
@@ -255,6 +255,8 @@ func tracerOptionsGCP(ctx context.Context) []trace.TracerProviderOption {
...
+		trace.WithSampler(trace.ParentBased(trace.TraceIDRatioBased(0.1))),
```
k4leung4, line 259: can we make this a var or env?

tcnghia: Done

k4leung4 (approved):
thank you

PR #458, opened 2024-07-22
handle eventing failure rate separately

k4leung4:
eventing failure rate does not justify being pageable
have it routed to slack only

### DIFF
--- a/modules/alerting/main.tf
+++ b/modules/alerting/main.tf
@@ -277,7 +277,7 @@ resource "google_monitoring_alert_policy" "service_failure_rate_eventing" {
 
   documentation {
     // variables reference: https://cloud.google.com/monitoring/alerts/doc-variables#doc-vars
-    subject = "Cloud Run service $${metric_or_resource.labels.service_name} had 5xx error rate above ${var.failure_rate_ratio_threshold} for ${var.failure_rate_duration}s"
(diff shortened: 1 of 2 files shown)


modules/alerting/main.tf
```diff
@@ -240,6 +240,56 @@ resource "google_monitoring_alert_policy" "service_failure_rate" {
...
+    }
+
+    display_name = "eventing services 5xx failure rate above ${var.failure_rate_ratio_threshold}"
+  }
+
+  display_name = "eventing services 5xx failure rate above ${var.failure_rate_ratio_threshold}"
+
+  documentation {
+    // variables reference: https://cloud.google.com/monitoring/alerts/doc-variables#doc-vars
+    subject = "Cloud Run service $${metric_or_resource.labels.service_name} had 5xx error rate above ${var.failure_rate_ratio_threshold} for ${var.failure_rate_duration}s"
```
cmdpdx, line 280:
```suggestion
    subject = "Eventing service $${metric_or_resource.labels.service_name} had 5xx error rate above ${var.failure_rate_ratio_threshold} for ${var.failure_rate_duration}s"
```
should we make this distinction? (maybe using more accurate language than I suggested)

cmdpdx (approved):

modules/alerting/main.tf
k4leung4: the service should be included, which should be sufficient.
naming it eventing should reinforce that.

PR #462, opened 2024-07-24
Remove laser alerts

tcnghia:


modules/secret/main.tf
```diff
@@ -39,49 +39,3 @@ resource "google_secret_manager_secret_iam_binding" "authorize-version-adder" {
 // Get a project number for this project ID.
 data "google_project" "project" { project_id = var.project_id }
 
-// Create an alert policy to notify if the secret is accessed by an unauthorized entity.
```
cmdpdx, line 42: Are there any we would consider keeping? If so, secret access seems like a candidate.

modules/serverless-gclb/main.tf
```diff
@@ -149,45 +149,3 @@ locals {
   )
 }
 
-resource "google_monitoring_alert_policy" "abnormal-gclb-access" {
```
k4leung4, line 152: should we keep this one, we dont have that many instances of serverless-gclb, it wont save much?

modules/secret/main.tf
k4leung4: agree this might have been one that matt wanted to keep

tcnghia: many of the "secrets" are in fact ConfigMap so probably not work keeping them as libraries. if we want to keep them I'd say we have to be explicit in an "alerting" module instead of a resource module where people may just accidentally create alerts in their dev env without knowing why

modules/configmap/main.tf
```diff
@@ -28,55 +28,3 @@ data "google_project" "project" { project_id = var.project_id }
 // What identity is deploying this?
 data "google_client_openid_userinfo" "me" {}
 
-// Create an alert policy to notify if the secret is accessed by an unauthorized entity.
```
k4leung4, line 31: secret related, maybe keep?

tcnghia: this is modules/configmap, so secret isn't really secret

modules/secret/main.tf
cmdpdx: what about an opt in var to the sercrets module?

modules/configmap/main.tf
tcnghia: we have so many of these in dev envs and it will multiply

k4leung4: you sure its not a secret? line 17 of this file is a secret in the secret manager

for dev, i think we want a knob to enable to turn on these alert policies, off by default, turn on for staging and prod

k4leung4:
i think we want a knob for alert policies that is off by default, and only staging/prod would turn on as to avoid fees in dev env

modules/configmap/main.tf
cmdpdx: IIRC using secrets here is an implementation detail, they are not considered to actually be secret. Those would use the secret module

modules/secret/main.tf
tcnghia: I reverted this module

modules/serverless-gclb/main.tf
tcnghia: This is the GCLB used in dev envs right? I don't think keeping it has the effect on non-dev/test envs

modules/configmap/main.tf
tcnghia: Yes, it is a "secret" because there is no ConfigMap object in CloudRun, and we have to mount using Secret. But both the module and the alert says "abnormal ConfigMap access"

mattmoor: Yeah, ok to kill `configmap` secret stuff.

modules/serverless-gclb/main.tf
mattmoor: Agree with Nghia that this isn't what we use in prod, so little benefit to keep.  If we want it, we should add it to the mono copy.

k4leung4: ack, remove it

mattmoor (approved):

PR #469, opened 2024-07-29
tune spammy alerts to only alert once a day

k4leung4:
add log based metrics to track frequency over time

### DIFF
--- a/modules/alerting/main.tf
+++ b/modules/alerting/main.tf
@@ -290,13 +290,32 @@ resource "google_monitoring_alert_policy" "service_failure_rate_eventing" {
   project = var.project_id
 }
 
+resource "google_logging_metric" "cloud-run-scaling-failure" {
+  name   = "cloud_run_scaling_failure"
+  filter = <<EOT
+        resource.type="cloud_run_revision"
+        log_name="projects/prod-enforce-fabc/logs/run.googleapis.com%2Frequests"
+        severity=ERROR
+        textPayload:"The request was aborted because there was no available instance."
+      EOT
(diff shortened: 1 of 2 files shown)


cpanato (approved):

modules/alerting/main.tf
```diff
@@ -290,13 +290,32 @@ resource "google_monitoring_alert_policy" "service_failure_rate_eventing" {
   project = var.project_id
 }
 
+resource "google_logging_metric" "cloud-run-scaling-failure" {
+  name   = "cloud_run_scaling_failure"
+  filter = <<EOT
+        resource.type="cloud_run_revision"
+        log_name="projects/prod-enforce-fabc/logs/run.googleapis.com%2Frequests"
```
imjasonh, line 297: Should this reference `var.project_id`?

modules/alerting/main.tf
```diff
@@ -328,13 +347,32 @@ resource "google_monitoring_alert_policy" "cloud-run-scaling-failure" {
   project = var.project_id
 }
 
+resource "google_logging_metric" "cloud-run-failed-req" {
+  name   = "cloud_run_failed_req"
+  filter = <<EOT
+        resource.type="cloud_run_revision"
+        log_name="projects/prod-enforce-fabc/logs/run.googleapis.com%2Frequests"
```
imjasonh, line 354: here too

modules/alerting/main.tf
k4leung4: nice catch
fixed.

PR #474, opened 2024-07-31
bot sdk: add method to clone repo

imjasonh:
This lets us reuse the octosts token and its refresh/revoke lifecycle, instead of having to manage it separately.


k4leung4 (approved):

modules/github-bots/sdk/github.go
```diff
@@ -452,3 +457,60 @@ func validateResponse(ctx context.Context, err error, resp *github.Response, act
...
+
+	// git clone <repo>
+	// git fetch origin <ref>:<ref>
+	// git checkout <ref>
+	r, err := git.PlainCloneContext(ctx, destDir, false, &git.CloneOptions{
+		URL: repo,
+		Auth: &gitHttp.BasicAuth{
+			Username: "notchecked", // username is not checked, only the token in the password field is used.
+			Password: tok.AccessToken,
+		},
```
ajayk, line 487: can we move this out 

```
    auth := &gitHttp.BasicAuth{
        Username: "notchecked", // username is not checked, only the token in the password field is used.
        Password: tok.AccessToken,
    }
```

imjasonh: Good call, done.

ajayk (approved):

PR #479, opened 2024-08-05
grant service account telemetry permissions for cron job service account

k4leung4:
mirrors that of regional service

https://github.com/chainguard-dev/terraform-infra-common/blob/main/modules/regional-service/main.tf#L1-L17


cmdpdx (approved):
Thank you!

PR #481, opened 2024-08-06
metrics sidecar for cloudrun job

k4leung4:
prometheus metrics are scrapped
otel sidecar does scrapping for prometheus metrics and push to gcp

still need to verify if it works
but i did verify that it doesnt break things

Screenshot 2024-08-06 at 2 23 05 PM

### DIFF
--- a/modules/cron/otel-config/config.yaml
+++ b/modules/cron/otel-config/config.yaml
@@ -1,0 +1,83 @@
+receivers:
+  prometheus:
+    config:
+      scrape_configs:
+      - job_name: "localhost"
+        scrape_interval: 10s
+        static_configs:
+        # TODO: make this configurable
+        - targets: ["localhost:2112"]
+        # Do not relabel job and instance labels if existed.
+        honor_labels: true
+        metric_relabel_configs:
+          - source_labels: [ __name__ ]
+            regex: '^prometheus_.*'
+            action: drop
+          - source_labels: [ __name__ ]
+            regex: '^process_.*'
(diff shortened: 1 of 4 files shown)


imjasonh (approved):

modules/cron/otel-config/config.yaml
```diff
@@ -0,0 +1,83 @@
...
+      from_attribute: faas.id
+      action: upsert
+    # The `gcp` resourcedetection processor sets `faas.name` to the name of the
+    # Cloud Run service or the Cloud Run job.
+    - from_attribute: faas.name
+      # The googlemanagedprometheus exporter consumes `service.name` attribute
+      # and set the `job` resource label to this value. (See
+      # https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/764)
+      key: "service.name"
+      action: upsert
```
cmdpdx, line 55: I'm not familiar with this syntax, but is it an issue if the fields are in different orders here? 

e.g. `- key:...` and then `- from_attribute: ...`

cmdpdx (approved):

modules/cron/otel-config/config.yaml
k4leung4: it doesnt matter

each entry as noted with `-` is a separate object in the `attributes` list, which is then parsed as an object where ordering of fields doesnt matter

cmdpdx: great, thanks!

PR #489, opened 2024-08-09
github-bots: remove trigger from the name to allow more custom chars

rawlingsj:
I tried creating a bot with a brief descriptive name, for example `my-team-update-check`  but got an error during the apply:

```
(""bot-trigger-my-team-update-check-bj0e") doesn't match regexp "^[a-z](?:[-a-z0-9]{4,28}[a-z0-9])$"
```

`bot-trigger-` is always added as a prefix, which makes it pretty easy to go over the 28 char regex restriction.

Could we remove the `trigger-` part?

### DIFF
--- a/modules/github-bots/main.tf
+++ b/modules/github-bots/main.tf
@@ -49,7 +49,7 @@ module "cloudevent-trigger" {
   source   = "../cloudevent-trigger"
 
   project_id                = var.project_id
-  name                      = "bot-trigger-${var.name}"
+  name                      = "bot-${var.name}"
   broker                    = var.broker[each.key]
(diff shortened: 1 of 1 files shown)


rawlingsj:
Having said this, I'd imagine this would recreate service account names made by the bot, which recreates their unique ID and break sts policies.  Hmm, ok let's not do this :)

imjasonh:
> Having said this, I'd imagine this would recreate service account names made by the bot, which recreates their unique ID and break sts policies.  Hmm, ok let's not do this :)

Does it? If not I think it's worth it.

PR #491, opened 2024-08-09
fix envconfig structs, bump clog to release

imjasonh:
Broken by #488


jdolitsky (approved):

tcnghia:
This still doesn't work as intended: I think Seth's syntax must be 
```
     Var type `env:"VAR", required`
```
instead of
```
     Var type `env:"VAR", required: "true"`
```

I tried this locally and `required: "true"` will turn required fields to optional.

tcnghia:
See https://github.com/sethvargo/go-envconfig?tab=readme-ov-file#configuration

PR #494, opened 2024-08-14
Add task_count and parallelism to cron module

tcnghia:


k4leung4 (approved):

modules/cron/variables.tf
```diff
@@ -89,6 +89,22 @@ variable "memory" {
...
+  default     = 1
+  description = "The number of tasks to run. "
+}
+
+variable "parallelism" {
+  type        = number
+  default     = 1
+  description = "The number of parallel jobs to run. Must be <= task_count"
+  validation {
+    condition     = var.parallelism <= var.task_count
```
cpanato, line 103: we cannot reference another variable in the condition, we need to create an object, and then we can test

PR #503, opened 2024-08-19
remove replace dep for trace 

k4leung4:
function in main line now

https://github.com/googleapis/google-cloud-go/issues/4665#issuecomment-2297676212

bump to go 1.23

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -15,7 +15,7 @@ permissions:
 jobs:
 
   test:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/go-test.yaml
```diff
@@ -15,7 +15,7 @@ permissions:
 jobs:
 
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-8-core
```
cmdpdx, line 18:
```suggestion
    runs-on: ubuntu-latest-8-cores
```

cmdpdx (approved):

PR #528, opened 2024-09-03
make lasers optional, off by default

k4leung4:


modules/bucket-events/main.tf
```diff
@@ -55,6 +55,22 @@ resource "google_service_account_iam_binding" "allow-pubsub-to-mint-tokens" {
   members = ["serviceAccount:${google_project_service_identity.pubsub.email}"]
 }
 
+module "audit-delivery-serviceaccount" {
+  count = var.enable_lasers ? 1 : 0
```
mattmoor, line 59: I wouldn't make this it's own flag.  I would key off of whether any notification channels have been passed and push it all the way in to the alert policy level.

modules/audit-serviceaccount/main.tf
```diff
@@ -0,0 +1,41 @@
...
+
+      label_extractors = {
+        "email"       = "EXTRACT(protoPayload.authenticationInfo.principalEmail)"
+        "method_name" = "EXTRACT(protoPayload.methodName)"
+        "user_agent"  = "REGEXP_EXTRACT(protoPayload.requestMetadata.callerSuppliedUserAgent, \"(\\\\S+)\")"
+      }
+    }
+  }
+
+  notification_channels = var.notification_channels
```
mattmoor, line 37: See my other comment, but it's there resources I'd make have a `count` and I'd predicate it on `len(var.notification_channels) > 0 ? 1 : 0`

modules/bucket-events/main.tf
k4leung4: are you suggesting this for all modules or just this one?
for other modules, there are other alert policies that notification channels can notify on
and so it doesnt make sense to turn off all notification to turn off lasers

mattmoor: I'm suggesting this for all alert policies.

If we don't pass in notification channels, then don't create the alert.

We don't pass notifications channels at all in dev environments, and as we bring up prod components we cycle thru none -> slack/email -> pagerduty.

I think "no channels" -> "no alert" makes sense, but we may be talking past each other, so may be helpful to talk through on a quick meet.

k4leung4: fixed.

modules/audit-serviceaccount/main.tf
k4leung4: done

k4leung4:
ptal

PR #534, opened 2024-09-05
tune down auto-close for logs based alerts

k4leung4:
logs based alerts doesnt re-trigger, it triggers a new instance
set these alerts to auto-close after an hour to avoid cluttering active alert list

### DIFF
--- a/modules/alerting/main.tf
+++ b/modules/alerting/main.tf
@@ -363,12 +363,12 @@ resource "google_logging_metric" "cloud-run-scaling-failure" {
 }
 
 resource "google_monitoring_alert_policy" "cloud-run-scaling-failure" {
-  # In the absence of data, incident will auto-close after an daily
+  # In the absence of data, incident will auto-close after an hour
   alert_strategy {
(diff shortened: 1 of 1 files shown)


modules/alerting/main.tf
```diff
@@ -365,10 +365,10 @@ resource "google_logging_metric" "cloud-run-scaling-failure" {
 resource "google_monitoring_alert_policy" "cloud-run-scaling-failure" {
   # In the absence of data, incident will auto-close after an daily
   alert_strategy {
-    auto_close = "86400s"
+    auto_close = "3600s"
 
     notification_rate_limit {
-      period = "86400s" // re-alert daily if condition still valid.
+      period = "3600s" // re-alert daily if condition still valid.
```
cmdpdx, line 371: super nit: should this be `hourly` instead of `daily` now?

cmdpdx (approved):

modules/alerting/main.tf
k4leung4: thanks, fixed.

PR #544, opened 2024-09-14
Migrate the go-workqueue libraries into common infra

mattmoor:
This is mostly copy/pasta and `sed` import rewrites thus far.


pkg/workqueue/workqueue.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+
+// This is actually github.com/chainguard-dev/terraform-infra-common/pkg/workqueue but the codegen
+// doesn't translate properly.
+option go_package = "github.com/chainguard-dev/workqueue";
```
mattmoor, line 5: Ah this is wrong now...

.github/workflows/go-test.yaml
```diff
@@ -47,7 +47,19 @@ jobs:
     - name: Test
       run: |
         # Exclude running unit tests against third_party repos.
-        go test -race ./...
+        go test -race -timeout=5m ./...
+
+    - run: |
+        cd "$(mktemp -d)"
+        wget -O protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v27.1/protoc-27.1-linux-x86_64.zip
```
k4leung4, line 54: nit:
can you have the version match those of the chainguard sdk

```
We currently require protoc v5.28.0.

Install protoc codegen dependencies:

go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.34.2
go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.5.1
go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway@v2.22.0
go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2@v2.22.0
```

https://github.com/chainguard-dev/mono/blob/main/sdk/README.md

mattmoor: `brew` pulled in 28.1, so I went with that (since downgrading is a PITA).

mattmoor: If we could make it stop putting precise versions in the header comments, I'd wager we could make our lives much much easier.

k4leung4: that would be nice

pkg/workqueue/conformance/conformance.go
k4leung4, line 1: is dedupe part of conformance?

i might have missed it, but i dont see any testing around that?

mattmoor: We could test it more explicitly, but it is covered by the check on ~142.

I think I'd rather follow up with adding that since this is basically a copy of what's upstream right now.

imjasonh (approved):

PR #598, opened 2024-10-16
github-bots: optionally use a pre-defined service account email

hectorj2f:
Allow passing a service account instead of always creating a service account.

### DIFF
--- a/modules/github-bots/main.tf
+++ b/modules/github-bots/main.tf
@@ -1,4 +1,5 @@
 resource "google_service_account" "sa" {
+  count        = var.service_account_email == "" ? 1 : 0
   account_id   = "bot-${var.name}"
   display_name = "Service Account for ${var.name}"
 }
@@ -9,7 +10,9 @@ module "service" {
   name            = var.name
   project_id      = var.project_id
   regions         = var.regions
-  service_account = google_service_account.sa.email
+
+  service_account = var.service_account_email == "" ? google_service_account.sa[0].email : var.service_account_email
+
 
   egress = "PRIVATE_RANGES_ONLY" // Makes GitHub API calls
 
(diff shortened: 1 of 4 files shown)


modules/github-bots/main.tf
```diff
@@ -9,7 +9,9 @@ module "service" {
   name            = var.name
   project_id      = var.project_id
   regions         = var.regions
-  service_account = google_service_account.sa.email
+
```
k4leung4, line 12: please change it such that when passing in a service account, have it not create the one here.

modules/github-bots/main.tf
```diff
@@ -1,4 +1,5 @@
 resource "google_service_account" "sa" {
+  count        = var.service_account_email == "" ? 1 : 0
```
k4leung4, line 2: as this changed the resource path, 
can you add a moved block
https://developer.hashicorp.com/terraform/language/modules/develop/refactoring#enabling-count-or-for_each-for-a-resource

```
moved {
  from = google_service_account.sa
  to = google_service_account.sa[0]
}
```

hectorj2f: Oh! nice!

rawlingsj:
I tested this patch against an existing bot + authorize service and the SA wasn't changed so AFAICT this won't affect existing deployments that use the default created SA.

rawlingsj:
@k4leung4 looks ok - you happy to merge?

modules/github-bots/variables.tf
```diff
@@ -123,3 +123,9 @@ variable "deletion_protection" {
   description = "Whether to enable delete protection for the service."
   default     = true
 }
+
+variable "service_account_email" {
+  description = "The email of the service account being authorized to invoke the private Cloud Run service."
```
imjasonh, line 128:
```suggestion
  description = "The email of the service account being authorized to invoke the private Cloud Run service. If empty, a service account will be created and used."
```

k4leung4 (approved):
lets try this
have a feeling this might not work

PR #608, opened 2024-10-23
default launch stage now that vpc access is GA

k4leung4:
this is to reduce no-op diffs

example

```
  # module.build.module.handler["extras"].module.this.google_cloud_run_v2_service.this["us-central1"] will be updated in-place
  ~ resource "google_cloud_run_v2_service" "this" {
        id                      = "projects/staging-enforce-cd1e/locations/us-central1/services/build-extras"
      ~ launch_stage            = "GA" -> "BETA"
```

ignored example;
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_run_v2_job#example-usage---cloudrunv2-job-limits

### DIFF
--- a/modules/cron/main.tf
+++ b/modules/cron/main.tf
@@ -163,6 +163,12 @@ resource "google_cloud_run_v2_job" "job" {
       }
     }
   }
+
+  lifecycle {
+    ignore_changes = [
+      launch_stage,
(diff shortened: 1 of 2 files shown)


cmdpdx (approved):
:nice:

PR #628, opened 2024-11-15
alerting: dont create metrics for squad alerts

k4leung4:
for https://github.com/chainguard-dev/internal-dev/issues/4886

this is to allow creating global alerts filtered by squads/teams

### DIFF
--- a/modules/alerting/main.tf
+++ b/modules/alerting/main.tf
@@ -426,7 +426,14 @@ resource "google_monitoring_alert_policy" "service_failure_rate_eventing" {
   project = var.project_id
 }
 
+moved {
+  from = google_logging_metric.cloud-run-scaling-failure
+  to   = google_logging_metric.cloud-run-scaling-failure[0]
+}
+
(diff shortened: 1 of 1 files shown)


modules/alerting/main.tf
```diff
@@ -672,7 +693,14 @@ resource "google_logging_metric" "cloudrun_timeout" {
   }
 }
 
+moved {
+  from = google_logging_metric.dockerhub_ratelimit
+  to   = google_logging_metric.dockerhub_ratelimit[0]
+}
+
 resource "google_logging_metric" "dockerhub_ratelimit" {
```
cmdpdx, line 701: dumb question: the rate limits are shared across all services, does it make sense to exclude those for squad labeled services?

k4leung4: yes, this is what this is doing

if squad field is empty, implies global, set count to 1, create it
if squad field is non-empty, implies squad alert, set count to 0, dont create

due to the oddity of using count, i need to move the old resource state to the new one with the index

PR #644, opened 2024-12-02
cloudevent-job: invokes a Job in response to an event

imjasonh:
This executes a Cloud Run Job in response to a triggered event. This works by setting up a simple invoker service which gets triggered, and calls the `jobs.execute` API.

TODO: envs, volumes, VPC (if we need it?), dashboards, use it and figure out what else I'm missing


modules/cloudevent-broker/variables.tf
```diff
@@ -60,3 +60,9 @@ variable "squad" {
     error_message = "squad needs to specified or disable check by setting require_squad = false"
   }
 }
+
+variable "otel_collector_image" {
+  type        = string
+  default     = "otel/opentelemetry-collector-contrib:latest"
```
cpanato, line 66: @k4leung4 is pinning to a specific digest see #645

modules/cloudevent-job/cmd/invoker/main.go
```diff
@@ -0,0 +1,92 @@
+/*
+Copyright 2023 Chainguard, Inc.
```
cpanato, line 2: nit

```suggestion
Copyright 2024 Chainguard, Inc.
```

modules/cron/variables.tf
```diff
@@ -185,7 +185,7 @@ variable "enable_otel_sidecar" {
 
 variable "otel_collector_image" {
   type        = string
-  default     = "chainguard/opentelemetry-collector-contrib:latest"
+  default     = "otel/opentelemetry-collector-contrib:latest"
```
cpanato, line 188: Is this expected?

PR #652, opened 2024-12-05
add toggle to only create global only

k4leung4:


cmdpdx (approved):
what situations are the intended use of this flag?

k4leung4:
> what situations are the intended use of this flag?

We currently have two usage of this module, global and one per squad. The issue with the global one is the notification channel receives all alerts. With the squad ones, they only get squad alerts. The idea is to turn off squad alerts in the global one, and inform squad/team to monitor their team or squad slack Ashley channel

cpanato (approved):

PR #655, opened 2024-12-11
github-events: Add installation, hook ID, and headers to event messages.

wlynch:
- Adds installation message for identifying GitHub App installations. These won't be present in the legacy repo based webhooks.
- Adds X-GitHub-Hook-ID as an extension to make it filterable by downstream clients.
- Records the incoming event headers in the body payload, which includes useful info for debugging (e.g. delivery IDs).

Part of https://github.com/chainguard-dev/internal-dev/issues/1513

### DIFF
--- a/modules/github-events/cmd/trampoline/main.go
+++ b/modules/github-events/cmd/trampoline/main.go
@@ -86,12 +86,20 @@ func main() {
 		event.SetSource(r.Host)
 		event.SetSubject(msg.Repository.FullName)
 		event.SetExtension("action", msg.Action)
+		// Needs to be an extension to be a filterable attribute.
+		// See https://github.com/chainguard-dev/terraform-infra-common/blob/main/pkg/pubsub/cloudevent.go
+		if id := r.Header.Get("X-GitHub-Hook-ID"); id != "" {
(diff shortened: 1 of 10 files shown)


modules/github-events/cmd/trampoline/main.go
```diff
@@ -86,12 +86,20 @@ func main() {
...
-			Body json.RawMessage `json:"body"`
+			When time.Time `json:"when"`
+			// See https://docs.github.com/en/webhooks/webhook-events-and-payloads#delivery-headers
+			Headers http.Header     `json:"headers"`
+			Body    json.RawMessage `json:"body"`
 		}{
-			When: time.Now(),
-			Body: payload,
+			When:    time.Now(),
+			Headers: r.Header,
```
imjasonh, line 101: Should this be added to the BQ schemas too?

k4leung4 (approved):

modules/github-events/cmd/trampoline/main.go
wlynch: Good call - added. Changed this to an explicit type so we don't need to worry about json handling in BQ.

cpanato (approved):

PR #658, opened 2024-12-11
support multiple notification channels for each type

k4leung4:


cpanato (approved):

k4leung4:
> is this a breaking change right?

it is, but the only user is likely just us, which i'll fix

PR #662, opened 2024-12-12
check sdk: return {Create,Update}CheckRunOptions

imjasonh:
Callers don't tend to work in `CheckRun`s themselves, but rather `CreateCheckRunOptions`es (to create runs) or `UpdateCheckRunOptions`es (to update them).

There's no built-in method to convert a `CheckRun` into either of those, so we'll just do it ourselves.

To avoid duplication, `UpdateCheckRun()` just calls `CreateCheckRun()` and twiddles its values.


cpanato (approved):
just fix the lint

cpanato:
lgtm

PR #674, opened 2024-12-19
add alert for when a service is updated, but traffic is not routed to latest revision

k4leung4:
this will trigger on every release (update service)


cmdpdx (approved):
nice

PR #699, opened 2025-01-22
Disable NAT for GKE module

k4leung4:


imjasonh (approved):

joshrwolf (changes requested):
err wait, this triggers a cluster rebuild, the public node change we want was already made on the node_pool?

PR #703, opened 2025-01-27
add team/squad labels to dlq topic

k4leung4:
this would enable routing dlq alerts to the appropriate team

### DIFF
--- a/modules/cloudevent-trigger/variables.tf
+++ b/modules/cloudevent-trigger/variables.tf
@@ -109,3 +109,20 @@ variable "ack_deadline_seconds" {
   type        = number
   default     = 300
 }
+
+variable "require_team" {
+  description = "Whether to require team variable to be specified"
+  type        = bool
+  default     = true
+}
+
+variable "team" {
+  description = "team label to apply to the service."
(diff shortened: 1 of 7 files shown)


modules/cloudevent-trigger/variables.tf
```diff
@@ -109,3 +109,20 @@ variable "ack_deadline_seconds" {
   type        = number
   default     = 300
 }
+
+variable "require_team" {
+  description = "Whether to require team variable to be specified"
+  type        = bool
+  default     = true
```
cmdpdx, line 116: will making this default to true break any callers?

cmdpdx (approved):

modules/cloudevent-trigger/variables.tf
k4leung4: it would,
they would notice it when bumping the module version
which i plan on specifying it for our usage
if you are concerned, we can default it to false

cmdpdx: I think it's fine, especially if we want to enforce this (which I think we do).

PR #705, opened 2025-01-29
gke: pass extra variables to module

cmdpdx:
- `deletion_protection` for deleting the cluster
- `network_config` block in `pools` var to configure the node pool network configs. Defaults to the current values if not provided.


pnasrat (approved):
LGTM

PR #714, opened 2025-02-06
fix: panic with inner client

hectorj2f:
related to this thread in slack: https://chainguard-dev.slack.com/archives/C02SD39C6BW/p1738862337902599

### DIFF
--- a/modules/github-bots/sdk/github.go
+++ b/modules/github-bots/sdk/github.go
@@ -51,11 +51,6 @@ func NewGitHubClient(ctx context.Context, org, repo, policyName string, opts ...
 		opt(&client)
 	}
 
-	// if client has not been configured withOpts then create a new client
-	if client.inner == nil {
-		client.inner = github.NewClient(oauth2.NewClient(ctx, ts))
-	}
-
 	return client
 }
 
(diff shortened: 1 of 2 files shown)


modules/github-bots/sdk/github.go
```diff
@@ -40,6 +40,7 @@ func NewGitHubClient(ctx context.Context, org, repo, policyName string, opts ...
 	})
 
 	client := GitHubClient{
+		inner:   github.NewClient(oauth2.NewClient(ctx, ts)),
```
cmdpdx, line 43: 🤔 how is this different than initializing the client on line 55/56 if `client.inner` is nil?

IdlePhysicist (approved):

k4leung4 (commented):
can we add a unit test so it would have caught the panic?

modules/github-bots/sdk/github.go
cmdpdx: oh I see, if you use https://github.com/chainguard-dev/terraform-infra-common/blob/045407d7ae21dcda87a044293db21dc0ef22bb8b/modules/github-bots/sdk/github.go#L133
there is no client yet.

cmdpdx: I think with this change you can also remove the `if client.inner == nil` block below since it is now irrelevant

hectorj2f: Done! :)

hectorj2f:
@k4leung4 I've added a call to the function that paniced.

cpanato (approved):

PR #722, opened 2025-02-19
alert for non-zero exit code

k4leung4:


modules/alerting/main.tf
```diff
@@ -343,6 +343,59 @@ resource "google_monitoring_alert_policy" "fatal" {
   project = var.project_id
 }
 
+locals {
+  exit_filter = <<EOF
+resource.type="cloud_run_revision" OR resource.type="cloud_run_job"
+textPayload:"Container called exit("
+-textPayload="Container called exit(0)."
+${local.squad_log_filter}
+EOF
```
wlynch, line 352: Can we add a param for allowed exit codes? Otherwise this is going to be super noisy for elastic build jobs. Ideally we should update elastic builds to have a user-error exit code, this way we don't alert on a test failing that isn't caused by build infra.

k4leung4: agree, i just checked, we only use the default exitcode, (1), when things fail

at a minimum, we should change elastic build to not use 1

k4leung4: updated to ignore exit code 130-149, that elastic build now uses for package build failures.
ptal

cmdpdx: do we want the ignorable exit codes to be configurable? Or at least to pass in an exit code filter?

k4leung4: i think its going to be difficult to have it easily configurable, 
as it needs to take on gcp log regex syntax

so unless we see more need to make it configurable, im inclined to leave it as is

PR #725, opened 2025-02-21
upgrade go to minimum 1.24

cpanato:


.github/workflows/style.yaml
```diff
@@ -71,15 +71,15 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@3041bf56c941b39c61721a86cd11f3bb1338122a # v5.2.0
         with:
-          go-version-file: './go.mod'
+          go-version: '1.24'
```
k4leung4, line 74: do we need to update this if go.mod is being updated?

cpanato: no

PR #733, opened 2025-02-26
cron module to support NFS volumes

tcnghia:
Similar to how regional-service does it https://github.com/chainguard-dev/terraform-infra-common/blob/main/modules/regional-service/variables.tf#L154-L158 https://github.com/chainguard-dev/terraform-infra-common/blob/main/modules/regional-service/main.tf#L306-L320


k4leung4:
please update terraform docs

k4leung4 (approved):

PR #734, opened 2025-02-26
enable cron module to run non-Go images

tcnghia:
Currently the cron module will perform a `ko` build atop a given base image. This PR expands it to make `importpath` optional, in which case we just run the base image itself.

This is useful in case we want to use `cron` module with `apko`, to customize the command or the `run-as` user.

### DIFF
--- a/modules/cron/variables.tf
+++ b/modules/cron/variables.tf
@@ -35,7 +35,8 @@ variable "service_account" {
 
 variable "importpath" {
   type        = string
-  description = "The import path that contains the cron application."
+  description = "The import path that contains the cron application. Leave empty to run the unmodified base image as the application."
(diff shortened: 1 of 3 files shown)


modules/cron/variables.tf
```diff
@@ -35,7 +35,8 @@ variable "service_account" {
 
 variable "importpath" {
   type        = string
-  description = "The import path that contains the cron application."
+  description = "The import path that contains the cron application. Leave empty to run the unmodified base image as the application."
```
k4leung4, line 38: can you clarify when this is desired.

tcnghia: Done

k4leung4 (approved):

PR #747, opened 2025-03-13
Add configurable Cloud Run Job failure alert policy

rawlingsj:
This commit adds a new optional alert policy for Cloud Run Jobs that triggers when job executions fail. Key changes:

- Added failure_alert_enabled variable (defaults to false)
- Added configuration variables for alignment period and duration
- Uses condition_threshold for immediate alerting on failures
- Keeps existing success alert policy unchanged

The failure alert is opt-in and complements the existing success alert by providing immediate notification when jobs fail rather than waiting for expected successes.


rawlingsj:
Marked as draft to gather some feedback and while I test

PR #751, opened 2025-03-15
feat(sdk): new GitAuth method on GitHubClient

luhring:
I found myself wanting to use the `go-git` library in the same context as using this `GitHubClient`.

The `GitHubClient` seems awesome as a centralized, elegant way to handle OctoSTS auth, that automatically does all the right things.

But then I wanted to do more git things to the repo. For example, I wanted to clone a repo, but I didn't want to do all of the specific maneuvers that the `GitHubClient`'s existing `CloneRepo` method does. I wanted the power of the raw `go-git` library, but I didn't want to have to reinvent an auth solution when I do that.

Let me know if there's a better solution to this problem!

### DIFF
--- a/modules/github-bots/sdk/github.go
+++ b/modules/github-bots/sdk/github.go
@@ -209,6 +209,16 @@ func (c GitHubClient) GitAuth() (transport.AuthMethod, error) {
 	return auth, nil
 }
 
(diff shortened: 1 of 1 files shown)


luhring:
Also added a `RepoURL` method, along the same train of thought.

vaikas (approved):

PR #780, opened 2025-04-19
feat: Add pull trigger abstraction to mirror push trigger.

vaikas:
Create the abstractions for creating a Trigger that does not push events, but rather makes them available for pulling via a PubSub subscription.

### DIFF
--- a/modules/cloudevent-pull-trigger/README.md
+++ b/modules/cloudevent-pull-trigger/README.md
@@ -1,0 +1,122 @@
+# `cloudevent-pull-trigger`
+
+```mermaid
+flowchart LR
+    subgraph "regional network"
+    A[[Pub/Sub topic]] -- creates and inserts --> B(Subscription)
+    B(Subscription) -- pulls --> C(Consumer)
+    end
+```
+
+This module abstracts regionalized event-triggered services, but assumes the
+pull based consumption of events. This is very similar to the
+[`cloudevent-trigger`](./../cloudevent-trigger/), but better suited for very
+bursty traffic, where you can control the consumption of events by how you
+choose to pull them from the subscription. It's intended to be used with the
+"Broker" abstraction is described by the sibling
+[`cloudevent-broker`](./../cloudevent-broker/) module.
+
+The intended usage of this module for consuming events is something like this:
+
+```hcl
+// Create a network with several regional subnets
+module "networking" {
+  source = "chainguard-dev/common/infra//modules/networking"
+
+  name       = "my-networking"
+  project_id = var.project_id
+  regions    = [...]
+}
+
+// Create the Broker abstraction.
+module "cloudevent-broker" {
+  source = "chainguard-dev/common/infra//modules/cloudevent-broker"
+
+  name       = "my-broker"
+  project_id = var.project_id
+  regions    = module.networking.regional-networks
+}
+
+// Set up regionalized triggers to deliver filtered events from our regionalized
+// brokers to subscriptions.
+module "cloudevent-pull-trigger" {
+  for_each = module.networking.regional-networks
+
+  source = "chainguard-dev/common/infra//modules/cloudevent-pull-trigger"
+
+  name       = "bar"
+  project_id = var.project_id
+  broker     = module.cloudevent-broker.broker[each.key]
+
+  // Only trigger on bar-type events.
+  filter     = { "type" : "dev.chainguard.bar" }
+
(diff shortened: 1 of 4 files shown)


tcnghia (approved):

modules/cloudevent-pull-trigger/README.md
```diff
@@ -0,0 +1,122 @@
...
+  // Only trigger on bar-type events.
+  filter     = { "type" : "dev.chainguard.bar" }
+
+  // Only trigger on events with subject matching "foo/*"
+  filter_prefix = { "subject": "foo/" }
+}
+
+// Grant your subscriber the permissions to subscribe to trigger created above.
+resource "google_pubsub_subscription_iam_member" "pull_event_subscriber" {
+  subscription = module.cloudevent-pull-trigger.pull-subscription
```
mattmoor, line 60: I think this needs a `for_each`?

modules/cloudevent-pull-trigger/README.md
```diff
@@ -0,0 +1,122 @@
...
+  role         = "roles/pubsub.subscriber"
+  member       = "serviceAccount:${var.cve_detection_sa_email}"
```
mattmoor, line 62: I wonder if we want an abstraction like `authorize-pull-subscription` that handles this similar to: https://github.com/chainguard-dev/terraform-infra-common/tree/main/modules/authorize-private-service

I did this (instead of just plumbing the URL around) to encourage folks to use this to look up the Cloud Run URLs in a way that ensured things were authorized.  There is a dual here for the service accessing the pull subscription.

modules/cloudevent-pull-trigger/main.tf
```diff
@@ -0,0 +1,90 @@
...
+
+  expiration_policy {
+    ttl = "86400s"
+  }
+
+  enable_message_ordering = true
+}
+
+// Configure the subscription to deliver the events matching our filter to this service
+// using the above identity to authorize the delivery..
```
mattmoor, line 68: We're not using the identity, leftover comment?

modules/cloudevent-pull-trigger/main.tf
```diff
@@ -0,0 +1,90 @@
...
+
+locals {
+  // See https://cloud.google.com/pubsub/docs/subscription-message-filter#filtering_syntax
+  filter-elements = concat(
+    [for key, value in var.filter : "attributes.ce-${key}=\"${value}\""],
+    [for key, value in var.filter_prefix : "hasPrefix(attributes.ce-${key}, \"${value}\")"],
+    [for key in var.filter_has_attributes : "attributes:ce-${key}"],
+    [for key in var.filter_not_has_attributes : "NOT attributes:ce-${key}"],
+  )
+}
```
mattmoor, line 30: I wonder if we should create an internal module to share this logic so that things don't diverge 🤔

modules/cloudevent-pull-trigger/variables.tf
```diff
@@ -0,0 +1,128 @@
...
+  default     = []
+}
+
+variable "private-service" {
+  description = "The private cloud run service that is subscribing to these events."
+  type = object({
+    name   = string
+    region = string
+  })
+}
```
mattmoor, line 82: Delete?

modules/cloudevent-pull-trigger/main.tf
```diff
@@ -0,0 +1,90 @@
...
+  upper   = false
+  special = false
+}
+
+// A dedicated service account for this subscription.
+resource "google_service_account" "this" {
+  project = var.project_id
+
+  account_id   = "${var.name}-${random_string.suffix.result}"
+  display_name = "Delivery account for ${var.private-service.name} in ${var.private-service.region}."
```
mattmoor, line 12: I think this is going to break?

mattmoor:
I think you need to add the new directory to the `.github/workflows` for validation and whatnot...

modules/cloudevent-pull-trigger/README.md
vaikas: don't think so.

modules/cloudevent-pull-trigger/README.md
vaikas: It's only for the service account, seems cleaner to just do this since you create this likely from the same place as you create the subscription. We can talk about this more, but using it like this seems easy/simple enough.

PR #783, opened 2025-04-21
add validation for `success_alert_alignment_period_seconds`

cmdpdx:
The alignment period + duration of a request can't be greater than 25h. We hardcode the duration to `alignment_period/4`, which caps how big `alignment_period` can be to `20h`. Add validation on this var to avoid pushing incorrect values unknowningly.


mgreau (approved):
👍

PR #785, opened 2025-04-21
Fix links to wiki documentation

mgreau:
The Eng Wiki has recently been restructured.

This commit updates links to the new wiki documentation URLs


cmdpdx (commented):
I'd perhaps argue we shouldn't have any links to the wiki considering this is a public repo. Could/should we just inline the playbook steps here?

mgreau:
> I'd perhaps argue we shouldn't have any links to the wiki considering this is a public repo. Could/should we just inline the playbook steps here?

good catch! I did not realize that when noticing the wrong link. Will update it

PR #790, opened 2025-04-24
add option to create a GCS sink for a DLQ pubsub subscription

cpanato:
we sometimes want the messages to arrive in a gcs bucket to check them.

Adding this option

cc @rawlingsj


modules/cloudevent-pull-trigger/main.tf
```diff
@@ -41,19 +41,51 @@ resource "google_pubsub_topic_iam_binding" "allow-pubsub-to-send-to-dead-letter"
...
-  enable_message_ordering = true
+  dynamic "cloud_storage_config" {
+    for_each = var.enable_dlq_bucket == false ? [] : [1]
+    content {
+      bucket = google_storage_bucket.dlq_bucket.name
+
+      max_duration = "300s" # 5 minutes
+      max_messages = 1000
+    }
+  }
```
rawlingsj, line 88: I think this can be simplified and avoid the dynamic content, I also just had to look up what the `max_duration` and `max_messages` did so maybe a comment for these would also help, assuming we don't want these configurable.

I.e. something like...

```
cloud_storage_config {
  # When enable_dlq_bucket is false, bucket will be null, which disables
  # the entire Cloud Storage export configuration regardless of other settings
  bucket = var.enable_dlq_bucket ? google_storage_bucket.dlq_bucket.name : null

  # Cloud Storage Export Configuration:
  # (Only applies when enable_dlq_bucket is true)
  #
  # These parameters determine when message batches are exported to Cloud Storage:
  #
  # max_duration = "300s" (5 minutes):
  # - Maximum time window before messages are exported, even if max_messages isn't reached
  # - Ensures timely processing even with low message volume
  #
  # max_messages = 1000:
  # - Maximum number of messages to batch before triggering an export
  # - Optimizes storage operations by reducing the frequency of writes
  #
  # Export occurs when either condition is met first.
  max_duration = "300s" # 5 minutes
  max_messages = 1000
}
```

wdyt?

vaikas:
@cpanato are you going to be able to update this, or what's the plan here?

cpanato:
> @cpanato are you going to be able to update this, or what's the plan here?

working on that

modules/cloudevent-pull-trigger/main.tf
```diff
@@ -41,19 +41,51 @@ resource "google_pubsub_topic_iam_binding" "allow-pubsub-to-send-to-dead-letter"
   members = ["serviceAccount:${google_project_service_identity.pubsub.email}"]
 }
 
+resource "google_storage_bucket" "dlq_bucket" {
```
vaikas, line 44: Formatting seems a bit off here.

vaikas: The spaces and such, if you run `tf fmt` it doesn't complain about anything? Like L56 below, I'd expect it to complain:
```
bucket =google_storage_bucket.dlq_bucket[count.index].name
```

because there's no space after the `=`, but maybe I'm just wrong.

cpanato: ah ok, i will check, but that is old code, I have a new one in my local, was running some tests to validate things before pushing it

cpanato:
@vaikas @rawlingsj PTAL

tested the terraform code to create the pubsub topic and subscription with and without the gcs bucket all working

cpanato:
off course, forgot to update the the variables :(

cpanato:
PTAL @vaikas @rawlingsj

PR #791, opened 2025-04-25
Alert only when Cloud Run Job exceeds retries

tcnghia:


modules/alerting/main.tf
```diff
@@ -366,11 +366,10 @@ resource "google_monitoring_alert_policy" "fatal" {
 locals {
   # ignore exit 0 and 130-149 (used by build job failures)
```
tcnghia, line 367:
```suggestion
```

imjasonh (approved):

PR #796, opened 2025-04-28
Add trace restoration to workqueue

tcnghia:


pkg/workqueue/gcs/gcs.go
```diff
@@ -79,6 +80,14 @@ func (w *wq) Queue(ctx context.Context, key string, opts workqueue.Options) erro
 	}
 	writer.Metadata[notBeforeMetadataKey] = opts.NotBefore.UTC().Format(time.RFC3339)
 
+	// Extract trace context from the context and store it in metadata
+	if span := trace.SpanFromContext(ctx); span.SpanContext().IsValid() {
+		traceparent := span.SpanContext().TraceID().String() + "-" +
```
k4leung4, line 85: nit: can you use a helper method for constructing the traceparent string, so we can add unit test for them, to see when a traceparent string would look like?

pkg/workqueue/gcs/gcs.go
```diff
@@ -381,8 +390,48 @@ func (o *inProgressKey) Context() context.Context {
...
 }
 
+// restoreTraceContext attempts to restore trace context from a traceparent in metadata
+func restoreTraceContext(ctx context.Context, metadata map[string]string) context.Context {
+	if tp, ok := metadata["traceparent"]; ok && tp != "" {
+		// Extract the trace components.
+		// We can't do anything if we got a broken traceparent, so we
+		// only set the trace context if we can parse it.
+		parts := strings.Split(tp, "-")
+		if len(parts) == 3 {
```
k4leung4, line 400: nit: move parsing traceparent into helper method to make this easier to read

PR #798, opened 2025-04-29
panic_filter: match text regex 'runtime/panic'

hectorj2f:
I propose a change to this alert to also handle runtime panics as part of the condition that fires this alert. I am not so sure we want to fire this alert even if the severity is NOT at the ERROR level. But I'd consider awkward to log panics as just info/warn log entries.

related to: https://chainguard-dev.slack.com/archives/C05CM0DM7RV/p1745920565303639

### DIFF
--- a/modules/alerting/main.tf
+++ b/modules/alerting/main.tf
@@ -199,8 +199,8 @@ resource "google_monitoring_alert_policy" "signal" {
 locals {
   panic_filter = <<EOF
 resource.type="cloud_run_revision" OR resource.type="cloud_run_job"
-severity=ERROR
-textPayload=~"panic: .*"
+severity>=DEFAULT
+textPayload=~"panic: .*|runtime/panic"
 ${var.panic_filter}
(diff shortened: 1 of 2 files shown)


modules/alerting/main.tf
```diff
@@ -199,8 +199,8 @@ resource "google_monitoring_alert_policy" "signal" {
 locals {
   panic_filter = <<EOF
 resource.type="cloud_run_revision" OR resource.type="cloud_run_job"
-severity=ERROR
-textPayload=~"panic: .*"
+severity<=ERROR
+textPayload=~"panic: .*|runtime/panic"
```
imjasonh, line 203:
```suggestion
textPayload=~"panic: .*" OR textPayload=~"runtime/panic"
```

Small thing, but I never trust regexes when there's a real boolean operator available.

hectorj2f:
@imjasonh This is ready for another look 🙏🏻 !

imjasonh (approved):

PR #801, opened 2025-04-30
github-events: Optimize JSON unmarshalling in webhook handler

rawlingsj:
Reduce webhook processing overhead by unmarshalling the JSON payload once instead of four times.

This change improves performance by:

  - Parsing the payload into a generic map at the beginning
  - Update helper functions that work with already parsed data
  - Maintaining full backward compatibility
  - Preserving all existing functionality for PR and merged detection

All tests pass with the optimized implementation.

### DIFF
--- a/modules/github-events/internal/trampoline/server.go
+++ b/modules/github-events/internal/trampoline/server.go
@@ -78,16 +78,22 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	// Unmarshal payload to extract necessary information
+	var parsedPayload map[string]interface{}
+	if err := json.Unmarshal(payload, &parsedPayload); err != nil {
+		log.Warnf("failed to unmarshal payload, cloud event headers will not be set: %v", err)
+	}
+
+	// Extract action from the payload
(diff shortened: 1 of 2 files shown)


modules/github-events/internal/trampoline/server.go
```diff
@@ -78,16 +78,22 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	// Unmarshal payload to extract necessary information
+	var parsedPayload map[string]interface{}
```
imjasonh, line 82: Could we make this the actual struct with fields we know we need?

`map[string]interface{}` is a bit harder to use (casting stuff to `string` like one line 88`) and it means we store all the deserialized data, not just the parts we need.

I normally wouldn't care so much about optimizing stuff like this, but this service is implicated by _every PR event_, and PR events can be pretty huge. If we're not careful we may wake up in a month to find this service requires a ton of memory and replicas.

rawlingsj: sounds good to me!

imjasonh (approved):
Thank you!

PR #810, opened 2025-05-06
cloud-event-recorder: retry the blob upload errors

tcnghia:


k4leung4 (commented):
unit test?

pkg/rotate/blob.go
```diff
@@ -111,10 +119,20 @@ func (u *uploader) Run(ctx context.Context) error {
...
-				return fmt.Errorf("failed to close blob file: %s %w", fileName, err)
+			// retry this 3 times
+			for i := 0; i < MaxUploadAttempt; i++ {
+				if err := writer.Close(); err == nil {
+					break
+				}
+				if i == MaxUploadAttempt-1 {
+					// last attempt, no more retry
+					return fmt.Errorf("failed to close blob file: %s %w", fileName, err)
+				} else {
```
k4leung4, line 130: else not needed?

tcnghia:
> unit test?

`bucket.Writer` is a concrete type from GCS Go library, and while they have memory-backed fakes, they have no mocks that I could use to simulate this network failure.

k4leung4 (approved):

PR #814, opened 2025-05-07
alert policy for cloud run job failures

k4leung4:
split existing exit code alert policy into two, one for service and one for jobs

service will use log message with exit code
job will use audit log for execution run to tell if it failed or not

a separate alert policy will be created for the cloud run job where we customize the exit code to differentiate between expected failures and system failures


tcnghia (approved):

cpanato (approved):
thanks

PR #815, opened 2025-05-07
feat: add cloudsql-postgres module

blancadesal:
This module provisions a Cloud SQL for PostgreSQL instance with production-ready defaults and convenient toggles for high-availability, automated backups, point-in-time recovery (PITR) and read-replicas.

### DIFF
--- a/modules/cloudsql-postgres/docs/auth.md
+++ b/modules/cloudsql-postgres/docs/auth.md
@@ -1,0 +1,50 @@
+# Authentication Model for Cloud SQL PostgreSQL Module
+
+This document explains **how workloads authenticate to Cloud SQL instances provisioned by the `cloudsql‑postgres` Terraform module**, covering both GKE and non‑Kubernetes consumers. The design relies on **Cloud SQL IAM database authentication**.
+
+## Summary
+
+* **IAM database authentication** is enabled on every instance via the `cloudsql.iam_authentication` flag ([Cloud SQL IAM auth guide](https://cloud.google.com/sql/docs/postgres/iam-authentication)).
+* **Workload Identity** binds a Kubernetes Service Account (KSA) to a Google Service Account (GSA) that has `roles/cloudsql.client`; the GSA's short‑lived IAM token becomes the database credential ([GKE Workload Identity overview](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity)).
+* Workloads can connect through the **Cloud SQL Auth Proxy** sidecar; this is the fully supported approach ([Auth Proxy v2 documentation](https://cloud.google.com/sql/docs/postgres/connect-auth-proxy)).
+* The module **never creates password users**; teams who need one must add a separate `google_sql_user` resource ([Managing PostgreSQL users](https://cloud.google.com/sql/docs/postgres/create-manage-users)).
+
+## 1. Connection options
+
+### 1.1. Cloud SQL Auth Proxy sidecar in GKE *(recommended)*
+
+1. **Bind KSA → GSA** with Workload Identity Federation ([binding guide](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#binding_ksa_to_gsa)).
+2. Module grants the bound **GSA** `roles/cloudsql.client` ([Cloud SQL IAM roles reference](https://cloud.google.com/sql/docs/postgres/iam-roles-permissions)).
+3. Pod runs the Auth Proxy v2 sidecar:
+
+   ```yaml
+   - name: cloud-sql-proxy
+     image: gcr.io/cloud-sql-connectors/cloud-sql-proxy:2
+     args: ["--private-ip", "--auto-iam-authn", "$(INSTANCE)"]
+   ```
+
+   `--auto-iam-authn` instructs the proxy to issue a fresh IAM token for each connection ([Auth Proxy v2 documentation](https://cloud.google.com/sql/docs/postgres/connect-auth-proxy)).
+4. Application connects to `localhost:5432`; no credentials are stored in the container.
+
+### 1.2. Auth Proxy on non-GKE hosts
+
+```bash
+cloud-sql-proxy --private-ip --auto-iam-authn $INSTANCE &
+psql "host=127.0.0.1 user=iam:$GSA_EMAIL sslmode=disable"
+```
+
+This method requires that the host running the Cloud SQL Auth Proxy has direct private IP connectivity to the Cloud SQL instance, meaning it must be within the same VPC network or a network connected via VPC peering or VPN. It will **not work** from local environments unless a secure network connection (e.g., VPN or Cloud Interconnect) to the Google Cloud VPC network is established.
+
+The proxy obtains IAM authentication tokens automatically using the host's Application Default Credentials or the VM's assigned service account. ([Auth Proxy guide](https://cloud.google.com/sql/docs/postgres/connect-auth-proxy))
+
+## 2. FAQs
+
+**Q – Does IAM auth work for read replicas?**
+If the `cloudsql.iam_authentication` flag is **enabled on the primary**, Cloud SQL automatically enables it on any read replicas created afterwards. If it's **disabled on the primary**, replicas can't use IAM authentication and the flag cannot be turned on later ([Read‑replica creation guide](https://cloud.google.com/sql/docs/postgres/replication/create-replica)).
+
+**Q – Which editions support IAM auth?**
+Both **Enterprise** and **Enterprise Plus** editions support IAM authentication; Cloud SQL defaults to Enterprise Plus for PostgreSQL 16+ and Enterprise for earlier versions ([Edition overview](https://cloud.google.com/sql/docs/postgres/editions-intro)).
+
+## 3. Terraform IAM examples
+
+TODO
--- a/modules/cloudsql-postgres/main.tf
+++ b/modules/cloudsql-postgres/main.tf
@@ -1,0 +1,136 @@
+terraform {
+  required_providers {
+    google = {
+      source = "hashicorp/google"
+    }
+  }
+}
+
+# Labels
+
+locals {
+  default_labels = {
+    cloudsql = var.name
+  }
+
+  squad_label = var.squad != "" ? {
+    squad = var.squad
+    team  = var.squad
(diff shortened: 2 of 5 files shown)


modules/cloudsql-postgres/main.tf
```diff
@@ -0,0 +1,122 @@
...
+        name  = database_flags.key
+        value = database_flags.value
+      }
+    }
+
+    user_labels = local.merged_labels
+  }
+
+  lifecycle {
+    prevent_destroy = true
```
cpanato, line 79: should we make this a variable to be able to teardown?

modules/cloudsql-postgres/main.tf
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    user_labels = merge(local.merged_labels, {
+      replica = "true"
+      region  = each.value
+    })
+  }
+
+  lifecycle {
+    prevent_destroy = true
```
cpanato, line 110: same here

modules/cloudsql-postgres/main.tf
blancadesal: oops, no:

```
│ Error: Variables not allowed
│
│   on main.tf line 110, in resource "google_sql_database_instance" "replicas":
│  110:     prevent_destroy = var.prevent_destroy
│
│ Variables may not be used here.
```

we can move `deletion_protection` to a var but not this one

cmdpdx: maybe pass the entire `lifecycle` as a var and use a dynamic block? https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks

modules/cloudsql-postgres/main.tf
```diff
@@ -0,0 +1,122 @@
...
+    disk_size             = var.storage_gb
+    disk_autoresize       = true
+    disk_autoresize_limit = var.disk_autoresize_limit
+
+    ip_configuration {
+      ipv4_enabled    = false
+      private_network = var.network
+    }
+
+    backup_configuration {
```
k4leung4, line 47: `binary_log_enabled` must be enabled if it was to be a master instance

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#master_instance_name-1

if replicas is non-empty, this should be set to true

modules/cloudsql-postgres/variables.tf
```diff
@@ -0,0 +1,151 @@
...
+  }
+}
+
+# Engine & Capacity
+
+variable "database_version" {
+  description = "Cloud SQL engine version."
+  type        = string
+  default     = "POSTGRES_17"
+}
```
k4leung4, line 44: we should add a validation to this variable that the database_version is prefixed with `POSTGRES_` to ensure users dont accidentally run mysql/sql server with this module

https://developer.hashicorp.com/terraform/language/values/variables#custom-validation-rules

modules/cloudsql-postgres/variables.tf
```diff
@@ -0,0 +1,151 @@
...
+variable "network" {
+  description = "Self‑link or name of the VPC network used for private IP connectivity."
+  type        = string
+}
+
+variable "require_squad" {
+  description = "Whether the `squad` variable must be specified."
+  type        = bool
+  default     = true
+}
```
k4leung4, line 25: we should drop this and make it require

we only added this option for existing modules
new modules should just require it

modules/cloudsql-postgres/main.tf
```diff
@@ -0,0 +1,122 @@
...
+resource "google_sql_database_instance" "this" {
+  name    = var.name
+  project = var.project
+  region  = var.region
+
+  database_version    = var.database_version
+  deletion_protection = var.deletion_protection
+
+  settings {
+    tier                  = var.tier
```
k4leung4, line 35: should we add an option for `edition`
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#edition-1

modules/cloudsql-postgres/main.tf
```diff
@@ -0,0 +1,122 @@
...
+    availability_type     = var.enable_high_availability ? "REGIONAL" : "ZONAL"
+    disk_type             = "PD_SSD"
+    disk_size             = var.storage_gb
+    disk_autoresize       = true
+    disk_autoresize_limit = var.disk_autoresize_limit
```
k4leung4, line 40: if we do `disk_`autoresize = true`, i dont think we want to give user the option to specify disk size

```
Note that this value will override the resizing from disk_autoresize if that feature is enabled. To avoid this, set lifecycle.ignore_changes on this field.
```
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#disk_size-1

modules/cloudsql-postgres/main.tf
```diff
@@ -0,0 +1,122 @@
...
+
+    ip_configuration {
+      ipv4_enabled    = false
+      private_network = var.network
+    }
+
+    backup_configuration {
+      enabled                        = var.backup_enabled
+      start_time                     = var.backup_start_time
+      point_in_time_recovery_enabled = var.enable_point_in_time_recovery
```
k4leung4, line 50: documentation here suggests `point_in_time_recovery_enabled` should be set to `true` for postgres

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#availability_type-1

modules/cloudsql-postgres/main.tf
k4leung4, line 1: general question, 
do you plan on handling sql user in this module or separately?

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_user

and related, the auth mechanism, 

i have strong preference for IAM auth if we can make it work
https://cloud.google.com/sql/docs/postgres/iam-authentication

mysql iam had some limitations due to our usage, so it is only used for breakglass

modules/cloudsql-postgres/main.tf
blancadesal: I think we should because either way Google provider will choose a default, and it's nicer to be explicit.

> Note: If the database version for your instance is PostgreSQL 16 or later, then the default Cloud SQL for PostgreSQL edition is Cloud SQL Enterprise Plus edition. If the database version is earlier than PostgreSQL 16, then the default edition is Cloud SQL Enterprise edition.

https://cloud.google.com/sql/docs/postgres/create-instance

modules/cloudsql-postgres/main.tf
blancadesal: That doesn't seem possible either. I'm wondering if we should skip this altogether as in the end it's enough to just delete it or comment it out to disable it, meaning it's all client-side and doesn't provide any strong protection.
https://www.reddit.com/r/Terraform/comments/zf7qju/variablize_prevent_destroy/

https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle
my interpretation is that `lifecycle` is not just another "nested block" that can be generated; it’s
a meta-block interpreted by Terraform itself, before any provider logic runs, and so any args must be literals and not dynamically evaluated.

modules/cloudsql-postgres/main.tf
blancadesal: > general question,
> do you plan on handling sql user in this module or separately?

I'd like to keep the infra and access layers separate. I suggest we create a helper module (later) if we want a turn-key solution for user provisioning.

> mysql iam had some limitations due to our usage, so it is only used for breakglass

Do you mean our mysql auth is password-based other than for break-glass? 

I'm working on a doc explaining the auth model I'm envisioning with usage examples from GKE and elsewhere, we can continue the discussion there.

modules/cloudsql-postgres/main.tf
k4leung4: i vote for removing it completely

as this isnt strictly used for prod, if it cant be configurable, we shouldnt prevent users from using it in a manner that allows it from being destroyed

modules/cloudsql-postgres/main.tf
k4leung4: thanks, wished google had better docs.

modules/cloudsql-postgres/main.tf
k4leung4: can we do something smarter where if running with master/replicas, we check that these properties are set to true so?

modules/cloudsql-postgres/main.tf
k4leung4: ack. we can discuss the auth model in the doc

separating infra/access layers makes sense
since the `google_sql_user` resource takes the instance name as an input, can you make sure that the instance name is an output of this module

blancadesal: https://github.com/chainguard-dev/terraform-infra-common/blob/07c2f451513a28c995074e849ef446fe9bf6f8b6/modules/cloudsql-postgres/docs/auth.md

blancadesal:
So far, I've tested the following in my dev environment:
* Stand up a bare-bones instance (no HA, no read replicas)
* Add a database and user via terraform
* Connect using IAM via a bastion VM running cloud-sql-proxy with IAP tunneling from my local machine (uhh, that's a mouthful)

k4leung4 (approved):
minor comment

modules/cloudsql-postgres/docs/auth.md
```diff
@@ -0,0 +1,68 @@
...
+   TOKEN=$(gcloud auth print-access-token)
+   ```
+2. Connect using `psql`:
+
+   ```bash
+   psql "host=${PRIVATE_IP} sslmode=require \
+         user=iam:${GSA_EMAIL} password=${TOKEN} dbname=myapp"
+   ```
+
+   The `iam:` prefix tells Cloud SQL to validate the OAuth 2.0 token ([IAM login syntax](https://cloud.google.com/sql/docs/postgres/iam-logins)).
```
k4leung4, line 43: considering removing this, as we avoid using tokens whenever possible

if you do decide to keep it, consider hiding it under a collapsed section
https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections

and add that you need the role `roles/cloudsql.instanceUser`

blancadesal: removed it

PR #826, opened 2025-05-14
feat: add bastion module

blancadesal:
Generic jump-host (bastion) module. The primary use case is to provide controlled, break-glass access to resources within a VPC that do not expose a public IP.

Includes an option to install the Cloud SQL Auth Proxy for secure, IAM-based access to Cloud SQL instances, but this is disabled by default. Otherwise, it's generic by design and isn't tied to any specific backend or service.

Needed for https://github.com/chainguard-dev/internal-dev/issues/12794, https://github.com/chainguard-dev/internal-dev/issues/12795

### DIFF
--- a/modules/bastion/main.tf
+++ b/modules/bastion/main.tf
@@ -1,0 +1,277 @@
+/*
+Copyright 2025 Chainguard, Inc.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+resource "google_project_service" "os_config_api" {
+  project            = var.project_id
+  service            = "osconfig.googleapis.com"
+  disable_on_destroy = false
+}
+
+locals {
+  // Derive region from zone (e.g. us-central1-a -> us-central1)
+  region        = join("-", slice(split("-", var.zone), 0, 2))
+  instance_name = substr(var.name, 0, 63) # GCE name limit
+  instance_tag  = var.name
+
+  // Labels
+  default_labels = {
+    bastion = var.name
+  }
+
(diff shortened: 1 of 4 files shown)


blancadesal:
This doesn't have the redis-specific bits yet (as I'd like to test that first), but I'd appreciate a **high-level review of the security and hardening** bits at this stage 🙏

modules/bastion/main.tf
```diff
@@ -0,0 +1,319 @@
...
+      augenrules --load
+    COMMON
+    ,
+
+    var.postgres_conn_name != "" ? <<-POSTGRES
+      # ------------------------------------------------------
+      # Install Cloud SQL Auth Proxy v2 (PostgreSQL)
+      # ------------------------------------------------------
+      curl -sSL -o /usr/local/bin/cloud-sql-proxy \
+        https://storage.googleapis.com/cloud-sql-connectors/cloud-sql-proxy/v2.8.2/cloud-sql-proxy.linux.amd64
```
cpanato, line 64: is this the latest?

how we will track the releases and update when needed?

cpanato (commented):
looks good so far, i think this is pretty cool
thanks for doing this

modules/bastion/main.tf
cpanato: but we can do in a follow up not a problem for this PR

blancadesal: it's v2.16 in the latest commit, this one is outdated

PR #851, opened 2025-05-27
Add in user-agent

joedborg:
This is a bit of a stab in the dark, so expecting revisions.

The goal here is to get user-agent added to our metrics that are exported to GCP so we can address https://github.com/chainguard-dev/internal-dev/issues/10510.  We need user-agent to filter out requests from `chainctl`.


pkg/httpmetrics/metrics.go
```diff
@@ -238,7 +238,10 @@ func tracerOptionsGCP(ctx context.Context) []trace.TracerProviderOption {
 	traceExporter, err := texporter.New(
 		// Avoid infinite recursion in trace uploads
 		//   https://github.com/open-telemetry/opentelemetry-go/issues/1928
-		texporter.WithTraceClientOptions([]option.ClientOption{option.WithTelemetryDisabled()}),
+		texporter.WithTraceClientOptions([]option.ClientOption{
+			option.WithTelemetryDisabled(),
+			option.WithUserAgent("otel-go"),
+		}),
```
k4leung4, line 244: this adds it to the tracer exporter, which isnt the http metrics routed to gcp monitoring

i think you want to update the prometheus counter metric here
https://github.com/chainguard-dev/terraform-infra-common/blob/main/pkg/httpmetrics/metrics.go#L86-L92

joedborg:
> > @k4leung4 also happy to cut the UA string to just the first segment (browser/app) if you think it's necessary?
> 
> i would prefer to cut it to the first two? browser/app and os i find full user agent strings too long and difficult to use
> 
> software type might be useful to identify bots? is there a user agent parser lib that we might be able to leverage

Sounds good!  Took a pass at rolling our own but I soon realised how much of an unstandard User-Agent headers are.  I've tested this, locally, with curl, Firefox, Chrome and chainctl.  Happy to play around with the particular metrics we're passing.

pkg/httpmetrics/metrics.go
```diff
@@ -300,11 +301,14 @@ func instrumentHandlerCounter(counter *prometheus.CounterVec, next http.Handler)
...
 		counter.With(prometheus.Labels{
-			"method":     r.Method,
-			"code":       strconv.Itoa(d.Status),
-			"ce_type":    r.Header.Get(CeTypeHeader),
-			"user_agent": r.Header.Get("User-Agent"),
+			"method":                     r.Method,
+			"code":                       strconv.Itoa(d.Status),
+			"ce_type":                    r.Header.Get(CeTypeHeader),
+			"user_agent_browser_name":    ua.Name,
+			"user_agent_browser_version": ua.Version,
```
k4leung4, line 310: do we want to use `VersionNoShort()` rather than the full version string to have better grouping?
https://github.com/mileusna/useragent?tab=readme-ov-file#version-parsing

and similarly for OS, OSVersionNoShort()

joedborg: Yep!  Happy to do that.  Pushing now.

PR #874, opened 2025-06-06
workqueue: Add error status details to not reenqueue errors.

wlynch:
Sometimes we don't want to re-enqueue and error because it's not retriable. This is something that's going to be very client specific, since we don't know under what conditions particular error codes may or may not be retried.

This change allows clients to attach additional details to the error, allowing them to inform the dispatcher when certain errors should not be reenqueued. This preservese the existing behavior where keys are reenqueued on error by default.


pkg/workqueue/dispatcher/dispatcher.go
```diff
@@ -114,6 +114,12 @@ func HandleAsync(ctx context.Context, wq workqueue.Interface, concurrency int, f
 					if err := oip.Deadletter(ctx); err != nil {
 						return fmt.Errorf("fail(after reaching max retries) = %w", err)
 					}
+				} else if d := workqueue.GetNonRetriableDetails(err); d != nil {
+					clog.InfoContextf(ctx, "Key %q is marked as non-retriable - reason: %s, err: %v", oip.Name(), d.GetMessage(), err)
+					// If the error is marked as non-retriable, we should not requeue it.
+					if err := oip.Complete(ctx); err != nil {
+						return fmt.Errorf("complete(after non-retriable error) = %w", err)
+					}
```
mattmoor, line 122: We need to be really careful about how we use this as we could shed some of the key resilience the workqueue gives us.

PR #875, opened 2025-06-06
add input variable `additional-service-account-accessors` to `module/secret`

cmdpdx:
There are cases where more than one service is required to access a secret value (e.g. when using a redis module with auth, any viewers will also need access to the auth string to make the connection).

Retain the `_binding` resource in the secret module, but allow additional accessors to be defined.

### DIFF
--- a/modules/secret/variables.tf
+++ b/modules/secret/variables.tf
@@ -13,14 +13,22 @@ variable "authorized-adder" {
 }
 
 variable "service-account" {
-  description = "The email of the service account that will access the secret."
+  description = "(Deprecated: Use service-accounts instead) The email of the service account that will access the secret."
   type        = string
+  default = ""
 }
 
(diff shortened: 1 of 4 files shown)


cpanato (dismissed):

modules/secret/variables.tf
```diff
@@ -17,6 +17,12 @@ variable "service-account" {
   type        = string
 }
 
+variable "additional-service-account-accessors" {
+  description = "Additional service accounts that will need to access the secret."
+  type = list(string)
+  default = []
+}
+
```
k4leung4, line 25: should we name this `service-accounts`, and note `service-account` is deprecated?

k4leung4 (dismissed):

mgreau (dismissed):

mgreau (approved):

PR #884, opened 2025-06-11
specify resource type for prometheus dashboard

k4leung4:


joedborg (approved):
Adding the 
` "resource.type=\"prometheus_target\"",`
filter, looks good to me.

PR #894, opened 2025-06-17
Add a github reconciliation "sdk" of sorts

mattmoor:


mattmoor:
I've deployed this to my dev environment with the changes here: https://github.com/chainguard-dev/mono/pull/24912

modules/github-workqueue/README.md
```diff
@@ -0,0 +1,90 @@
+# GitHub Workqueue Module
+
+This Terraform module provisions a GitHub webhook service that receives GitHub events and enqueues them to a workqueue for processing. It acts as a bridge between GitHub webhooks and the workqueue service, allowing you to process GitHub events asynchronously.
+
+## Architecture
+
+The module creates a **Webhook Service** that:
+- Receives GitHub webhook events via HTTP (through a load balancer)
+- Validates webhook signatures for security
```
vaikas, line 9: Looks like we create a random secret in the module, would be good to document how to get access to it so you can wire it up on the GitHub side. At least it used to be the case that you also needed to go twiddle GH to wire the secret up, but maybe that's not the case anymore? Or put another way, is there any additional stuff that one needs to do besides running this TF, or does this "Just Work" ™️ once you launch it?

pkg/githubreconciler/client.go
```diff
@@ -0,0 +1,92 @@
...
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/chainguard-dev/clog"
+	"github.com/google/go-github/v72/github"
+	"golang.org/x/oauth2"
+)
+
+// TokenSourceFunc is a function that creates an OAuth2 token source for a given org/repo.
```
vaikas, line 18: Is there a reason we can't use this instead of creating a new one, or update that one?
https://github.com/chainguard-dev/terraform-infra-common/blob/c13ca4656a52fc90160e8134c2d4cabf234173cb/modules/github-bots/sdk/github.go#L36

The sprawl of various GH clients seems

mattmoor: Yeah, I originally built this in mono and moved it over, so this can be combined.

Generally the stuff in `modules/` is specific to the binaries in that module, so it probably makes sense to move this stuff into `pkg/`.

I'll get this cleaned up when I have a break

mattmoor: done, in a second commit to keep things a bit cleaner

modules/github-workqueue/cmd/shim/main.go
```diff
@@ -0,0 +1,270 @@
...
+	_ "github.com/chainguard-dev/clog/gcp/init"
+	"github.com/chainguard-dev/terraform-infra-common/pkg/workqueue"
+	"github.com/google/go-github/v72/github"
+	"github.com/sethvargo/go-envconfig"
+)
+
+type Config struct {
+	Port             string `env:"PORT,default=8080"`
+	WorkqueueService string `env:"WORKQUEUE_SERVICE,required"`
+	WebhookSecret    string `env:"GITHUB_WEBHOOK_SECRET,required"`
```
imjasonh, line 24: We already have github-events, which handles webhook requests and puts them into PubSub. Could this be a more general PubSub->Workqueue, or at least GitHub Events->Workqueue?

I could see us wanting GCS Events->Workqueue, and in general Events->Workqueue eventually, even soon, and being able to reuse as much of the existing Events infra would be nice.

Also, not needing to set up a second webhook endpoint for everything seems a lot nicer from an operational perspective.

pkg/githubreconciler/client.go
```diff
@@ -0,0 +1,92 @@
...
+	clog.FromContext(ctx).With(
+		"org", org,
+		"repo", repo,
+	).Info("Created new GitHub client for repository")
+
+	return client, nil
+}
+
+// Clear removes all cached clients.
+func (cc *ClientCache) Clear() {
```
imjasonh, line 88: Does anything besides tests call `Clear`? I'm not sure I understand when this would be used in practice.

pkg/githubreconciler/reconciler.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/google/go-github/v72/github"
+)
+
+// ReconcilerFunc is the function signature for GitHub resource reconcilers.
+// It receives the parsed resource information and appropriate GitHub client,
+// and returns an error if reconciliation fails.
+type ReconcilerFunc func(ctx context.Context, res *Resource, gh *github.Client) error
+
+// Resource represents a parsed GitHub resource (issue or pull request).
```
imjasonh, line 20: There are lots of event types we might care about, e.g., pushes, and CheckRun events (retries, etc.)

We already support lots of those events in github-events, and have BQ recording set up with schemas. Can we reuse any of that here?

modules/github-workqueue/cmd/shim/main.go
mattmoor: This is several comments in one, but in most of them you are glossing over a lot of details that make a workqueue a workqueue: **we need to know the key from the payload**, and that isn't something that necessarily generalizes across heterogeneous events (even normalized to cloud events without some pretty extreme rigor I doubt we have, e.g. making the CE-Subject of everything relating to a PR the PR URL).

The notion of bridging the raw event stream to the Broker and having this consume events from the broker is definitely a potentially interesting way of reorienting this to avoid needing N webhooks for N reconcilers.  I was somewhat reluctant to do this out of the gate, but it's absolutely something we will want.

I think the decomposition that makes the most sense is:
1. A service for: `github -> GH flavored CloudEvents`
2. A service for `GH flavored CloudEvents -> workqueue`

This module is effectively the composition of these two things today, and it's behavior could be replicated by simply composing these two services directly.

Alternately, we could point `1.` at the Broker and consume `2.` from the Broker (to get what you want).

modules/github-workqueue/cmd/shim/main.go
```diff
@@ -0,0 +1,270 @@
...
+	ResourceFilter   string `env:"RESOURCE_FILTER"` // Optional: "issues" or "pull_requests"
```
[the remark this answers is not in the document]
mattmoor: I think we could express this as N filters (for each of the relevant event types), which I think practically speaking would manifest as N triggers (I'm not sure we're flexible enough to combine them today, or if PubSub is even)

modules/github-workqueue/cmd/shim/main.go
mattmoor: It's possible that `github-events` has what we need here, I can look.

pkg/githubreconciler/client.go
mattmoor: I suspect you are right.  I'll ask Claude what it was thinking 😉

pkg/githubreconciler/reconciler.go
mattmoor: See my comment above, but it's not such a simple thing to generically map events to workqueues.

Think about all of the helper functions that exist to do this mapping in K8s:
1. Watching our own resource type
2. Watching child resource types and chasing OwnerRefs
3. Leveraging things like the Knative `tracker` to watch ObjectReferences.
...

modules/github-workqueue/cmd/shim/main.go
imjasonh: > A service for `GH flavored CloudEvents -> workqueue`

If that's where this ends up, I'd be happy.

I do think there's a possibility of us defining some generalized way to map an Event to a Key, possibly even in some declarative language like CEL which we can embed in HCL, so that this generalizes to ~everywhere we subscribe to events -- e.g., for apk push events, the key is the apk name; for registry push events, you might want the key to be the repo:tag, or possibly something else. But I'm fine punting on that generally and seeing how it works for GH events specifically, to get more experience with it.

pkg/githubreconciler/reconciler.go
imjasonh: I buy that it's not easy to do in general, I just wanted to reuse as much as we have already rather than only building from scratch.

In a workqueue-centric model, do we not still want to record all the events, and get support for types we already have a use case for?

If we didn't compose the two, do you envision that we should separately subscribe to GH events to get them into BQ, and handle CheckRun retry webhooks via PubSub and issue updates via WQ? I hope not.

modules/github-workqueue/cmd/shim/main.go
mattmoor: Ok, so the `cloudevents-workqueue` implementation is [here](https://github.com/chainguard-dev/terraform-infra-common/pull/902).  I'll rebase this on that next and strip out the module here, which really just leaves the reconciler skeleton stuff

pkg/githubbot/github.go
```diff
@@ -1,4 +1,4 @@
-package sdk
+package githubbot
```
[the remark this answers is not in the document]
mattmoor: I undid this, so this shouldn't be breaking.

imjasonh (approved):
🤩

mattmoor:
@imjasonh I have a "review manager" I've been working on as well that works a lot like the "state manager" but clears prior reviews when posting new ones (all based on knobots prior art).  Will follow up with that tho.

PR #900, opened 2025-06-20
fix up dashboard filter

k4leung4:
previous dashboard relied on the `RESOURCE_LABEL` `service_name` metric, which is only available for built in google cloud run metrics.

to have it work for our custom prometheus metric, we add `METRIC_LABEL` `service_name`
now that we insert a metric `service_name` label for all cloud run services.
we can now leverage the dashboard filter to select the service rather than needing to specify it for each widget

as sections end up having a mix of build-in and custom prometheus metrics, module level filter with `resource.type=cloud_run_revision` no longer apply
add the resource type to each widget and remove the one at the module level

address issue where it dashboard does not filter on service
https://chainguard-dev.slack.com/archives/C083P5AQNQ2/p1750171310685599

### DIFF
--- a/modules/bucket-events/dashboard.tf
+++ b/modules/bucket-events/dashboard.tf
@@ -20,7 +20,7 @@ module "http" {
 module "resources" {
   source        = "../dashboard/sections/resources"
   title         = "Resources"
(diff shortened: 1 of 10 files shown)


joedborg (dismissed):
Looks like validation is failing but the code changes, in principal, lgtm!

cpanato (approved):
🚀

PR #910, opened 2025-06-23
update bastion

k4leung4:
require 2fa for login

install elastic defend 
copied from https://github.com/chainguard-dev/infra/blob/main/terraform/modules/bastion/main.tf#L120C1-L183C46

tweak to restarts


blancadesal (approved):
looks cool, thanks!

PR #915, opened 2025-06-25
replace pagerdutty with agnostic name for paging 

cpanato:
xref: https://github.com/chainguard-dev/mono/pull/25067

### DIFF
--- a/modules/alerting/variables.tf
+++ b/modules/alerting/variables.tf
@@ -8,8 +8,8 @@ variable "notification_channels" {
   default     = []
 }
 
-variable "notification_channels_pagerduty" {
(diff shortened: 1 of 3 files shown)


modules/alerting/variables.tf
```diff
@@ -8,8 +8,8 @@ variable "notification_channels" {
   default     = []
 }
 
-variable "notification_channels_pagerduty" {
-  description = "Email notification channel."
+variable "notification_channels_incident_io" {
```
k4leung4, line 11:
```suggestion
variable "notification_channels_pager" {
```

lets rename it for the purpose and avoid using a service name

PR #917, opened 2025-06-25
Add SA impersonation var for module created GKE SA

joedborg:
Objective is to allow a consumer of the GKE module the ability to pass an email, belonging to another SA, to be bound to the SA created by the GKE module for impersonation.

### DIFF
--- a/modules/gke/main.tf
+++ b/modules/gke/main.tf
@@ -33,7 +33,7 @@ resource "google_project_iam_member" "cluster" {
 }
 
 resource "google_service_account_iam_member" "terraform_gke_impersonation" {
-  count              = var.service_account_impersonation_email == null ? 1 : 0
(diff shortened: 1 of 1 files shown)


modules/gke/variables.tf
```diff
@@ -22,13 +22,13 @@ variable "require_squad" {
 }
 
 variable "squad" {
-  description = "squad label to apply to the service."
+  description = "Squad label to apply to the service."
```
cpanato, line 25: Note: Should we replace the squad with team?

cpanato (approved):
seems good

modules/gke/variables.tf
joedborg: I'm happy to, in this PR, if that's the way we're going?

cpanato: i think that will require more things, thanks!

PR #918, opened 2025-06-25
clog: ensure gcp logging is enabled

imjasonh:
I noticed a few common services weren't optimizing their logs for GCP, which meant they were missing some details (trace ID, source location, correct severity)

This fixes that, anywhere I could find a `cmd/main.go` that looked like a service without the import.


mattmoor (approved):
Wondering if there's anything like `depcheck` that we could use to make sure we're throwing these 🤔

PR #923, opened 2025-06-26
Standard convention for labeling by module

jml:
- Follows up from https://github.com/chainguard-dev/terraform-infra-common/pull/913
- Pattern originally introduced in #511

Per feedback in https://github.com/chainguard-dev/terraform-infra-common/pull/913#discussion_r2167352412

I've added documentation as well.

If we think these labels aren't actively used, I'm also happy to remove them. 

If we want to go further into module-based labelling, then it would probably be useful to have a `module: <module-name>` label as well, rather than one where `<module-name>` is the key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,3 +17,30 @@ provider "google" {
   project = var.project
 }
 ```
+
+## Resource Labeling Convention
+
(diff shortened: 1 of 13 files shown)


k4leung4 (approved):
thank you

PR #934, opened 2025-07-03
clean up ci and add .go-version file

cpanato:
Add a .go-version file to set the go version there and not in the go modules, so as not to force downstream users to use the same go version if they don't want to.


jml (commented):
Makes sense to me. I'm not familiar enough with Go packaging. Is there a mechanism to say that the code here assumes Go >=1.24?

cpanato:
Today, we are configuring Go in CI using whatever is defined in the go.mod, but if we always want to use the latest available, that means we need to bump the go.mod as well. The drawback is that we force downstream consumers to be on that Go version and sometimes this is not something we need to.

jml (approved):

PR #942, opened 2025-07-09
regional-(go-)service: add require_auth

imjasonh:
Previously attempted in https://github.com/chainguard-dev/terraform-infra-common/pull/939

This adds an optional (default `false`) var to `regional-(go-)service` to require GCP IAM auth and `roles/run.invoker` to call the service.

Prior to this, there was no other way to get a `.run.app` service that required auth and wasn't publicly invokable.


k4leung4 (approved):
change is fine

can you comment on whats the use case?

PR #955, opened 2025-07-16
add module to be able to create widgets with promql

cpanato:


modules/dashboard/widgets/xy-promql/main.tf
```diff
@@ -0,0 +1,31 @@
...
+        timeSeriesQuery = {
+          prometheusQuery = var.promql_query
+        }
+      }]
+      thresholds = [
+        for threshold in var.thresholds : {
+          value      = threshold
+          targetAxis = "Y1"
+        }
+      ],
```
k4leung4, line 23: should we keep this as a list of threshold objects to keep it the same as the other widgets?

https://github.com/chainguard-dev/terraform-infra-common/blob/main/modules/dashboard/widgets/xy-ratio/main.tf#L50

k4leung4 (approved):

PR #964, opened 2025-07-21
httpmetrics: Warn once if buckets are not setup

markusthoemmes:
Instead of flooding the logs with all the hosts that aren't matching, this logs a warning about buckets not being setup once and then always returns "other".

### DIFF
--- a/pkg/httpmetrics/transport.go
+++ b/pkg/httpmetrics/transport.go
@@ -182,7 +182,16 @@ func instrumentRoundTripperDuration(next http.RoundTripper) promhttp.RoundTrippe
 	}
 }
 
+var setupWarning sync.Once
+
 func bucketize(ctx context.Context, host string) string {
+	if buckets == nil && bucketSuffixes == nil {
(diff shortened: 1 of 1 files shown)


pkg/httpmetrics/transport.go
```diff
@@ -182,7 +182,16 @@ func instrumentRoundTripperDuration(next http.RoundTripper) promhttp.RoundTrippe
 	}
 }
 
+var setupWarning sync.Once
+
 func bucketize(ctx context.Context, host string) string {
+	if buckets == nil && bucketSuffixes == nil {
+		setupWarning.Do(func() {
+			clog.WarnContext(ctx, "no buckets configured, use httpmetrics.SetBuckets or SetBucketSuffixes")
```
konradzapalowicz, line 190: love that this msg provides useful context on desired action to resolve problem

konradzapalowicz (approved):

k4leung4 (approved):
thank you

PR #969, opened 2025-07-22
dashboard: Fix typo and make label dynamic to the errors

markusthoemmes:
This fixes a small typo and makes the respective label dynamic to the list of response codes we want to ignore.


konradzapalowicz (approved):
neat

PR #980, opened 2025-07-24
regional-X-service: default to GEN2

imjasonh:
GEN2 is generally what we want, and I've been surprised more than twice lately that some service is running as GEN1.

I think we set GEN1 as the default for backward compatibility, and so we could slowly adopt GEN2, but I think we pretty much have everywhere, and even if we haven't I think we should.


k4leung4 (changes requested):
i would prefer to hold off on this until we can test it on a heavy loaded service before updating the default

PR #986, opened 2025-07-28
github-events: add extra debug level logging to list all cloud event extensions being set

rawlingsj:


cpanato (changes requested):
lgtm

need to re-generate the docs

rawlingsj:
@cpanato thanks, looks like it was a CI flake, retriggering made it green but the PR is blocked with the request for changes

PR #987, opened 2025-07-28
workqueue: Store last attempted timestamp, compute wait metrics based on that time.

wlynch:
Currently, we compute the wait time based off of the created time. This means that if a key is re-attempted we are computing wait times off of the original time, which is not reflective of the actual queue wait time between attempts.

This changes to add a last-attempted timestamp when the object is re-enqueued, and modifying the metrics to compute off of this one if present, otherwise falling back to the original create time.

Unix timestamp used to minimize storage size.


k4leung4 (approved):

k4leung4 (commented):
did you want to add a unit test for this?

PR #989, opened 2025-07-29
Add required attributes for read replicas

joedborg:
Attempts to fix the read replicas of Postgres (https://chainguard-dev.slack.com/archives/C07U75JPGF4/p1753734338191569)

I believe adding `master_instance_name` is enough but have added some attributes to prevent a read replica being used as a failover and to not back up the read replicas.

### DIFF
--- a/modules/cloudsql-postgres/main.tf
+++ b/modules/cloudsql-postgres/main.tf
@@ -98,17 +98,27 @@ resource "google_sql_database_instance" "this" {
 resource "google_sql_database_instance" "replicas" {
   for_each = toset(var.read_replica_regions)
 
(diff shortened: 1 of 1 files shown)


modules/cloudsql-postgres/main.tf
```diff
@@ -129,6 +139,8 @@ resource "google_sql_database_instance" "replicas" {
       settings[0].disk_size
     ]
   }
+
+  depends_on = [google_sql_database_instance.this]
```
cpanato, line 143: i like the depends_on but you dont need that as you make a dependency when you link that in the line 105

cpanato (approved):

PR #998, opened 2025-07-31
dynamic create the advanced_machine_features otherwise dont

cpanato:
For existing clusters that are using older versions of the module, these settings did not exist, and if we upgrade, it tries to replace the node pools, but at this point, we want to have a minimum disruptive change, so if the setting is not set, it will keep the old behavior.


k4leung4 (approved):
thanks

PR #1020, opened 2025-08-14
feat: set team/squad variable defaults to 'unknown'

jml:
## Summary

Set default values for `team` and `squad` variables to "unknown" across all Terraform modules.

## Impact

This lets us distinguish between:

- could be labeled with a team but we haven't set the value yet, and
- haven't done the work to make it label-able

## Follow-up

The "is required" variables are no longer meaningful. My plan is to do one pass to strip out the logic, then find all our call sites & remove it there, then remove it here and release as a breaking change.

### DIFF
--- a/modules/bucket-events/variables.tf
+++ b/modules/bucket-events/variables.tf
@@ -55,6 +55,7 @@ variable "deletion_protection" {
   default     = true
 }
 
+
 variable "require_squad" {
   description = "Whether to require squad variable to be specified"
   type        = bool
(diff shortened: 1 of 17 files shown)


cpanato (approved):

modules/bucket-events/variables.tf
```diff
@@ -55,6 +55,7 @@ variable "deletion_protection" {
   default     = true
 }
 
+
```
k4leung4, line 58: nit: why are we adding an extra blank line ?

its in a number of files, but not all

jml: The robot did it! 🤖 

I'll fix.

PR #1027, opened 2025-08-20
Updates on ci / terraform and go

cpanato:
- clean up ci
- add more modules for terraform docs check
- add terraform version file
- upgrade to go1.25


.github/workflows/documentation.yaml
```diff
@@ -15,28 +16,35 @@ jobs:
       fail-fast: false
       matrix:
         module:
```
k4leung4, line 18: can we automate the discovery of the folders some how
as this will go stale really quickly

PR #1034, opened 2025-08-21
dinamically discover modules to check the tf docs

cpanato:


.github/workflows/documentation.yaml
```diff
@@ -8,44 +8,59 @@ on:
...
+          persist-credentials: false
+
+      - name: Discover modules with README.md
+        id: set-matrix
+        run: |
+          # Find all directories under modules/ that contain both README.md and .tf files
+          MODULES=()
+          while IFS= read -r -d '' dir; do
+            # Check if the directory contains both README.md and at least one .tf file
+            if [[ -f "$dir/README.md" ]] && ls "$dir"/*.tf >/dev/null 2>&1; then
```
k4leung4, line 37: not sure if we need to check for README, as that assumes its in a good state

worried we would miss new modules if they didnt add a README.md

cpanato: should check only for .tf files?

PR #1045, opened 2025-08-27
Add a histogram metrics for workqueue retries

tcnghia:


pkg/workqueue/doc.go
```diff
@@ -1,11 +1,112 @@
 /*
-Copyright 2024 Chainguard, Inc.
+Copyright 2025 Chainguard, Inc.
```
k4leung4, line 2: should this really be updated?

pkg/workqueue/gcs/metrics.go
```diff
@@ -96,6 +96,14 @@ var (
 		},
 		[]string{"service_name", "revision_name"},
 	)
+	mRetryAttemptsDistribution = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "workqueue_retry_attempts",
+			Help:    "Distribution of retry attempts across all tasks.",
+			Buckets: []float64{1, 2, 5, 10, 20, 30, 40, 50, 100, 200, 500, 1000},
```
tcnghia, line 103:
```suggestion
			Buckets: []float64{1, 2, 5, 10, 20, 50, 100, 200, 500, 1000},
```

pkg/workqueue/doc.go
tcnghia: Reverted!

PR #1048, opened 2025-08-28
Add regional CPU billing control to regional service modules

jml:
## What

Add a new `regional-cpu-idle` variable to `containers` in `regional-service` and `regional-go-service`. This controls whether we're using instance or request based billing based on region. The all-regions setting still exists and overrides it.

## Why

A few of our Cloud Run services have an optimization recommendation to switch to instance-based billing, but only in some regions. All told, these savings are estimated at ~$1k/month. Not huge, but not nothing either.

If we want to take advantage of this, we need to control request vs instance based billing at a regional level.

## Examples

```hcl
   # Per-container regional CPU billing configuration
   containers = {
     "main" = {
       # Different billing strategies by region for this container
       regional-cpu-idle = {
         "us-central1"    = false  # Instance billing for steady primary region traffic
         "europe-west1"   = true   # Request billing for bursty failover region traffic
       }

       # Setting it in `resources` is possible, I guess for backwards compatibility or simplicity.
       resources = {
         cpu_idle = false  # This container always uses instance billing, overriding regional settings
       }
     }

     "sidecar" = {
       # Different container can have completely different regional policy
       regional-cpu-idle = {
         "us-central1" = true   # Always request billing for sidecars in primary region
       }
     }
   }
```


modules/regional-go-service/variables.tf
```diff
@@ -259,3 +259,9 @@ variable "product" {
   type        = string
   default     = "unknown"
 }
+
+variable "regional_cpu_idle" {
```
k4leung4, line 263: rather than having a standalone var, i think it should mirror how we do `regional-env`
https://github.com/chainguard-dev/terraform-infra-common/blob/main/modules/regional-go-service/variables.tf#L92-L95

jml: Thanks!

It's updated now, and the examples look like:

```hcl
   # Per-container regional CPU billing configuration
   containers = {
     "main" = {
       # Different billing strategies by region for this container
       regional-cpu-idle = {
         "us-central1"    = false  # Instance billing for steady primary region traffic
         "europe-west1"   = true   # Request billing for bursty failover region traffic
       }

       # Setting it in `resources` is possible, I guess for backwards compatibility or simplicity.
       resources = {
         cpu_idle = false  # This container always uses instance billing, overriding regional settings
       }
     }

     "sidecar" = {
       # Different container can have completely different regional policy
       regional-cpu-idle = {
         "us-central1" = true   # Always request billing for sidecars in primary region
       }
     }
   }
```

PR #1051, opened 2025-08-29
workqueue: add global scope to fix deduplication and single-worker gu…

mattmoor:
…arantees

The regional architecture (separate buckets per region) breaks two key workqueue properties:
- Deduplication: same key can exist in multiple regional buckets
- Single worker: multiple regional dispatchers can process same logical key

Add `scope` variable accepting "regional" (default) or "global". When global, all regions use a single multi-regional bucket to restore proper workqueue semantics.

### DIFF
--- a/modules/workqueue/dispatcher.tf
+++ b/modules/workqueue/dispatcher.tf
@@ -69,8 +69,12 @@ module "dispatcher-service" {
       ]
       regional-env = [
         {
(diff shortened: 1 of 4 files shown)


vaikas:
Could we add some documentation on what is the right way to hold this so it's not lost in the PR description.

PR #1052, opened 2025-08-29
Add a metric/chart for attempts at completion.

mattmoor:
The idea behind this it to try and help tune our DLQ with practical data about how many retries are needed for things that ACTUALLY end up succeeding.

If we make max-retry too high, then we just thrash, so hopefully this helps us tune things.

### DIFF
--- a/modules/workqueue/README.md
+++ b/modules/workqueue/README.md
@@ -120,6 +120,7 @@ No requirements.
 
 | Name | Source | Version |
 |------|--------|---------|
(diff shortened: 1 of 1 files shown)


mattmoor:
@k4leung4 this is worth an extra careful look.  Metrics/Charts aren't my forte :)

tcnghia:
> @tcnghia WDYT?

Since you have a dashboard too, let's merge yours.  If this isn't sufficient, I can add mine later :pray

PR #1054, opened 2025-09-01
feat!: Remove require_squad and require_team variables

jml:
Instead, `squad` and `team` default to `"unknown"`.

This is a breaking change.
We have removed all known uses of these two variables within Chainguard-owned repos.

Follows on from #1020


cpanato (approved):
should be a minor version when release it

cpanato:
> Why minor? I'd have thought that a breaking API change should be major?
We can debate here for days :)

Major, I see besides a breaking change, some new big features, and maybe a revamp of the thing.

Minor can include breaking changes, and as this is not a huge api change, I would fit in a minor bump

PR #1072, opened 2025-09-09
workqueue: Don't break metrics out by location if scope is global

markusthoemmes:
Some of these metrics are misleading if broken out per region. Thus this cuts over most metrics to no longer break out per-region statistics and instead aggregate across all of them.

Notably, this excludes "Work processing latency" and "Amount of work added" as these allow for spotting regional imbalances that would otherwise be ignored.


mattmoor (approved):
cc @tcnghia

tcnghia (approved):

PR #1079, opened 2025-09-10
variable to control local disk use for recorder

k4leung4:
preparing when we can start using local disk for cloud run

### DIFF
--- a/modules/cloudevent-recorder/recorder.tf
+++ b/modules/cloudevent-recorder/recorder.tf
@@ -78,8 +78,11 @@ module "this" {
     }
   }
   volumes = [{
-    name      = "logs"
-    empty_dir = {}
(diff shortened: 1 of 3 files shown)


modules/cloudevent-recorder/recorder.tf
```diff
@@ -78,8 +78,11 @@ module "this" {
     }
   }
   volumes = [{
-    name      = "logs"
-    empty_dir = {}
+    name = "logs"
+    empty_dir = {
+      medium = var.local_disk_mount ? "DISK" : "MEMORY"
+      size   = var.local_disk_mount ? "16G" : "2G"
```
mattmoor, line 84: Will this have the effect of dropping our current tmpfs size?

PR #1100, opened 2025-09-22
Prefer empty over `.github`

mattmoor:
These should be the same (I plan to fix octo sts), but right now if you use `.github` we try to parse it with a repo policy instead of an org policy and it fails.


mattmoor:
ref: https://github.com/octo-sts/app/pull/1055

PR #1105, opened 2025-09-23
gh-bots: allow callers to overwrite the default cloudevent trigger name prefix

cmdpdx:
The default `cloudevent-trigger` name prefix of `bots-trigger` often causes issues for callers due to the length restrictions of the GSA created in the module. This results in the need to shorten desired names into difficult to decipher strings. As an alternative, allow callers to define an alternate (shorter) name prefix for the `cloudevent-trigger` module embedded within `github-bots`.


cmdpdx:
> _To offer an alternative to yet more variables..._
> 
> What we did in knative when resources started to hit length limits was: assuming a limit of N, and a suffix of length M, if length(prefix) + M > N then we would truncate the prefix to `prefix[:N-M]` and just append the suffix.
> 
> I think we should probably introduce a module that takes:
> 
> * prefix
> * limit
> * suffix
> 
> ... and performs this dance, and we can feed and length-limited name creation through it.
> 
> The other benefit of this approach was: **everything keeps the same name, except for cases that would fail today, so it is fully backwards compatible.**

I like that idea (though that won't let me be rid of the `bot-trigger` prefix that irks me so 🤣 ). I'll take a crack at it and likely close this PR.

PR #1107, opened 2025-09-24
feat: cpu_idle for ingress, workqueue, reconciler

jml:
`cpu_idle = false` means instance-based billing, which is cheaper for us for some services in some regions.

- Part of https://github.com/chainguard-dev/internal-dev/issues/20110
- Related to https://github.com/chainguard-dev/mono/pull/27115

I've tested this in my dev environment.

### DIFF
--- a/modules/cloudevent-broker/ingress.tf
+++ b/modules/cloudevent-broker/ingress.tf
@@ -47,6 +47,7 @@ module "this" {
       resources = {
         limits = var.limits
       }
+      regional-cpu-idle = var.cpu_idle
     }
   }
 
--- a/modules/cloudevent-broker/README.md
+++ b/modules/cloudevent-broker/README.md
@@ -109,6 +109,7 @@ No requirements.
 
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
+| <a name="input_cpu_idle"></a> [cpu\_idle](#input\_cpu\_idle) | Set to false for a region in order to use instance-based billing. Defaults to true. | `map(bool)` | `{}` | no |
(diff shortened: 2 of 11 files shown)


jml:
> If we should just be setting this everywhere, should we just flip it on in regional-service and be done with it?
> 

We shouldn't be setting it everywhere, it varies both by service and region. https://github.com/chainguard-dev/mono/pull/27150 has tables that I put together from manually trawling Cloud Run Services in the console (although I just figured out how to do this on the CLI, will need some permissions & API enablement).

> I’d rather avoid knobs if we can and if we add them, have their default state be what we want folks using without needing to learn to reach for it.

Me too, but I think this has to be tuned. 

> For places like the workqueue, do we need to be able to toggle this independently for the receiver and dispatcher?

Looks that way. `enforce-prod-cs-wq-dsp` wants it set in `us-central1` and `us-west1` but not `us-east1`, and `apk-scan-wq-prod-dsp` wants it set only in `us-central1`.

jml:
> (although I just figured out how to do this on the CLI, will need some permissions & API enablement).

- https://github.com/chainguard-dev/mono/pull/27381
- https://github.com/chainguard-dev/infra/pull/1210

modules/cloudevent-broker/ingress.tf
```diff
@@ -47,6 +47,7 @@ module "this" {
       resources = {
         limits = var.limits
       }
+      regional-cpu-idle = var.cpu_idle
```
cpanato, line 50: what is the default value for this? i think the var.cpu_idel should match the default value

mattmoor (approved):

PR #1117, opened 2025-09-28
Fix non-intuitive `RequeueAfter(0)` behavior.

mattmoor:
Today the way we handle `RequeueAfter(0)` is to return a wrapped error with the value set to 0.

The way this is turned into a response in Process methods is to extract it and populate the value.

The problem is that populating a value of 0 is indistinguishable from not setting it, so things do not get requeued.

Instead, we are changing RequeueAfter with any value less than 1s (we use second-resolution for delays) will return a normal error to trigger an immediate retry.


jml:

PR #1122, opened 2025-09-30
github-bots/sdk: Don't immediately revoke token on first use

wlynch:
There's no need to revoke the token if it's still valid. Instead, reuse the existing token and carry on as normal, respecting the existing token expiration.

### DIFF
--- a/modules/github-bots/sdk/github.go
+++ b/modules/github-bots/sdk/github.go
@@ -34,14 +34,13 @@ import (
 // A new token is created for each client, and is not refreshed. It can be
 // revoked with Close.
 func NewGitHubClient(ctx context.Context, org, repo, policyName string, opts ...GitHubClientOption) GitHubClient {
-	ts := oauth2.ReuseTokenSource(nil, &tokenSource{
-		org:        org,
-		repo:       repo,
-		policyName: policyName,
-	})
+	ts, err := octosts.NewTokenSourceFromValues(ctx, policyName, org, repo)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


modules/github-bots/sdk/github.go
```diff
@@ -34,14 +34,13 @@ import (
...
 // revoked with Close.
 func NewGitHubClient(ctx context.Context, org, repo, policyName string, opts ...GitHubClientOption) GitHubClient {
-	ts := oauth2.ReuseTokenSource(nil, &tokenSource{
-		org:        org,
-		repo:       repo,
-		policyName: policyName,
-	})
+	ts, err := octosts.NewTokenSourceFromValues(ctx, policyName, org, repo)
+	if err != nil {
+		clog.FromContext(ctx).Warnf("failed to create octosts token source, failing back to no token source: %v", err)
```
mattmoor, line 39: This masks an error that's (very likely) just going to manifest downstream as 403s, and because the client is bad, it won't recover.  I don't think we should just log and continue here.

mattmoor (commented):
You need to rebase to pick up an unrelated linting fix.

modules/github-bots/sdk/github.go
wlynch: Okay, reverted the old transport back for this particular case (but not for the default happy path). FWIW - this wouldn't fail because of github/octosts quota - this would only fail if we failed to fetch the GCP OIDC token the first time.

jml:

PR #1130, opened 2025-10-01
add a module to concat two strings, subject to a given maximum length

cmdpdx:
There are a few resources that have constraints on the length of their names. We currently apply the same name to a bunch of subresources within modules, and if one violates a maximum length requirement the entire module fails. This module intends to allow us to only shorten names when necessary, and otherwise continue using the caller provided names as is.

### DIFF
--- a/modules/limited-concat/variables.tf
+++ b/modules/limited-concat/variables.tf
@@ -1,0 +1,19 @@
+variable "prefix" {
+  description = "First part of the result, will be shortened if length(prefix)+length(suffix) > limit."
+  type = string
+}
+
+variable "suffix" {
+  description = "Second part of the result, included in whole."
+  type = string
+}
+
+variable "limit" {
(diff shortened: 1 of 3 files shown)


modules/limited-concat/variables.tf
```diff
@@ -0,0 +1,30 @@
...
+variable "limited-parameter" {
+  description = "Which parameter (prefix of suffix) to shorten if concatenation is longer than the limit."
+  type = string
+  default = "prefix"
+
+  validation {
+    condition = var.limited-parameter == "prefix" || var.limited-parameter == "suffix"
+    error_message = "limited-parameter must be one of \"prefix\" or \"suffix\"."
+  }
+}
```
mattmoor, line 30: I wouldn't overcomplicated it :)

mattmoor (approved):

PR #1136, opened 2025-10-03
add module for slo, and make it optional, disabled by default for regional-service

k4leung4:
for https://github.com/chainguard-dev/internal-dev/issues/20392

new module to create monitoring service to be associated with slo monitoring
configurable for cloud run service.

first pass is to support only availability
future iteration will focus on per method and correctness of these methods

will plumb it through other modules once its sufficiently tested in staging


cpanato (approved):
love it

thanks!

PR #1149, opened 2025-10-07
not fail in apply when we not set the slo var object

cpanato:
failing when the `slo` is not defined

```
Plan: 2 to add, 1 to change, 1 to destroy.
╷
│ Error: Attempt to get attribute from null value
│ 
│   on .terraform/modules/gh_rate_limit_service/modules/regional-service/main.tf line 373, in module "slo":
│  373:   count = var.slo.enable ? 1 : 0
│     ├────────────────
│     │ var.slo is null
│ 
│ This value is null, so it does not have any attributes.
╵

```


egibs (approved):

cpanato:
we need a new release

PR #1150, opened 2025-10-07
Check whether var.slo is null before accessing enabled

egibs:
We're seeing errors like this since `var.slo` defaults to `null`:
```
Error: Attempt to get attribute from null value

  on .terraform/modules/.../modules/regional-service/main.tf line 373, in module "slo":
 373:   count = var.slo.enable ? 1 : 0
    ├────────────────
    │ var.slo is null

```

This PR checks whether `var.slo` is `null` before accessing the value of `enabled`.

### DIFF
--- a/modules/regional-service/main.tf
+++ b/modules/regional-service/main.tf
@@ -370,7 +370,7 @@ resource "google_cloud_run_v2_service_iam_member" "public-services-are-unauthent
 }
 
 module "slo" {
-  count = var.slo.enable ? 1 : 0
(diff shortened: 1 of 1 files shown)


cpanato:
@egibs did simiar in https://github.com/chainguard-dev/terraform-infra-common/pull/1149 this morning :)

egibs:
> @egibs did simiar in #1149 this morning :)

Nice 🙂 -- I'll close this out.

